PR #7, opened 2020-08-19
Moving util into a superclass for migrators to extend, with abstract …

mseaton:
…revert and migrate methods.


mseaton:
@mogoodrich and @brandones - I figure I should start putting changes in PRs so that we stay aligned.  I just started down this path after our meeting to try to handle a few loose ends - 1. the ability to customize the DB credentials (you can specify these in a properties file that is specified in an environment variable), and 2. a bit more of an abstraction idea around executing migrations (with a migrate and revert method).  I also wanted to test out the revert functionality (which works).

A few loose ends here - for now we'll need to comment out or not the revert line as needed, based on how this is being executed.  Will need to test the properties file environment variables.  I set this up to default to what was hard-coded before, so it still works for me without any properties file in an environment variable, but you guys should test out if this works (and we'll need to document it once established)

PR #8, opened 2020-08-20
Support step, revert, and limit command-line arguments

brandones:
This makes developing a single step easier and faster.

### DIFF
--- a/etl/src/main/java/org/pih/hivmigration/etl/sql/Migrator.java
+++ b/etl/src/main/java/org/pih/hivmigration/etl/sql/Migrator.java
@@ -4,20 +4,47 @@ import org.apache.commons.lang3.time.StopWatch;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
+import javax.sound.midi.SysexMessage;
+import java.lang.reflect.InvocationTargetException;
+
 /**
  * Main class for executing the HIV migration using Spark
  */
 public class Migrator {
 
     private static final Log log = LogFactory.getLog(Migrator.class);
+    private static boolean shouldRevert = true;
 
     /**
      * Run the application
      */
 	public static void main(String[] args) {
         log.info("Starting up HIV Migrator");
-        revert(new UserMigrator());
-        migrate(new UserMigrator());
+        String revertStr = System.getenv("REVERT");
+        shouldRevert = revertStr.equals("1");
+        String step = System.getenv("STEP");
+        if (step != null && !step.isEmpty()) {
+            String className = "org.pih.hivmigration.etl.sql." + step + "Migrator";
+            try {
+                Class cls = Class.forName(className);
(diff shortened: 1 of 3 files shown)


etl/src/main/java/org/pih/hivmigration/etl/sql/SqlMigrator.java
```diff
@@ -26,23 +26,32 @@
...
+
+    public void setRowLimit(int rowLimit) {
+        this.rowLimit = rowLimit;
+    }
+
     abstract void migrate();
     abstract void revert();
 
     Properties getMigrationProperties() {
-        Properties p = new Properties();
```
brandones, line 33: All I did was memoize this function, really just so it doesn't produce that log message on every step.

etl/src/main/java/org/pih/hivmigration/etl/sql/SqlMigrator.java
```diff
@@ -57,8 +66,11 @@ Properties getOracleConnectionProperties() {
     Properties getMysqlConnectionProperties() {
         Properties mp = getMigrationProperties();
         Properties properties = new Properties();
-        properties.put("url", mp.getProperty("mysql.url", "jdbc:mysql://localhost:3308/openmrs?rewriteBatchedStatements=true"));
+        properties.put("url", "jdbc:mysql://" + mp.getProperty("mysql.host", "localhost")
+                + ":" + mp.getProperty("mysql.port", "3308") + "/"
+                + mp.getProperty("mysql.database", "openmrs") + "?rewriteBatchedStatements=true");
```
brandones, line 71: These are the properties that were present in the existing `migration.properties` file

etl/src/main/java/org/pih/hivmigration/etl/sql/SqlMigrator.java
```diff
@@ -129,8 +145,13 @@ void loadFromOracleToMySql(String targetStatement, String sourceQuery) throws Ex
                                                 targetConnection.commit();
                                                 batchesProcessed++;
                                                 rowsToProcess = 0;
-                                                log.info("Rows committed: " + batchesProcessed * batchSize);
+                                                if (batchSize * batchesProcessed % 50000 < batchSize) {
```
brandones, line 148: Log approximately every 50000 instead of every batch

etl/src/main/java/org/pih/hivmigration/etl/sql/Migrator.java
```diff
@@ -1,27 +1,81 @@
...
+                .parse(args);
+        migrator.main();
+    }
+
+    public void main() {
+	    if (step != null && !step.isEmpty()) {
+	        log.info("Only running " + step + "Migrator.");
+        }
+	    if (shouldRevert) {
+	        log.info("Will revert before each step.");
```
mogoodrich, line 48: Nit, but based on what we were talking about re: revert order, this should probably just read "Will revert before running".

PR #12, opened 2020-08-26
UHM-4790: Migrate "Migrate Patients" to SQL Migrator

mogoodrich:


etl/src/main/java/org/pih/hivmigration/etl/sql/util/FileParser.java
```diff
@@ -0,0 +1,32 @@
...
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileReader;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+public class FileParser {
+
+    // TODO confirm that this works with CSVs with more than one column (if we ever need it)
```
mogoodrich, line 12: Yeah, bad practice to not even test at all, but from a quick review we may never need this beyond the zl-identifier.csv, which is basically one column

etl/src/main/java/org/pih/hivmigration/etl/sql/EncounterMigrator.groovy
```diff
@@ -105,6 +105,6 @@ class EncounterMigrator extends SqlMigrator {
 
     void revert() {
         clearTable("encounter")
-        executeMysql("DROP TABLE hivmigration_encounters;")
+        executeMysql("DROP TABLE if exists hivmigration_encounters;")
```
brandones, line 108: Oh yeah, let's do this for all of them

etl/src/main/java/org/pih/hivmigration/etl/sql/InfantMigrator.groovy
```diff
@@ -1,7 +1,5 @@
 package org.pih.hivmigration.etl.sql
 
-import org.w3c.dom.Text
```
brandones, line 3: Thanks!

etl/src/main/java/org/pih/hivmigration/etl/sql/InfantMigrator.groovy
```diff
@@ -131,12 +129,16 @@ class InfantMigrator extends SqlMigrator {
 
     @Override
     def void revert() {
-        executeMysql("Remove infants from Relationship, Patient, Person Name, Person tables", '''
+
+        if (tableExists("hivmigration_infants")) {
```
brandones, line 133: Oh nice, that's a good way of doing it

etl/src/main/java/org/pih/hivmigration/etl/sql/util/FileParser.java
mseaton: Fine, but might be worth just pulling in an existing tried and tested CSV library that does this out of the box - something like opencsv, etc.

mogoodrich: True.. I was considering that, but this seemed simple enough to write... worth exploring further if there are other csv files we end up needing to import.

etl/src/main/java/org/pih/hivmigration/etl/sql/InfantMigrator.groovy
mogoodrich: IntelliJ does it automatically, didn't even notice this!

PR #16, opened 2020-09-01
UHM-4796 Migrate "Create Visits" to SQL Migrator

brandones:
https://pihemr.atlassian.net/browse/UHM-4796

From https://wiki.openmrs.org/display/docs/How+to+create+visits+for+preexisting+encounters

### DIFF
--- a/etl/src/main/java/org/pih/hivmigration/etl/sql/VisitMigrator.groovy
+++ b/etl/src/main/java/org/pih/hivmigration/etl/sql/VisitMigrator.groovy
@@ -16,13 +16,14 @@ class VisitMigrator extends SqlMigrator {
                 (
                     SELECT patient_id,
                            Min(encounter_datetime) AS date_started,  -- encounter_datetime is always midnight
-                           Addtime(Max(encounter_datetime), '23:59:00') AS date_stopped,
+                           Addtime(Max(encounter_datetime), '23:59:59') AS date_stopped,
                            Max(location_id) as location_id,  /* Avoid using 'Unknown Location' if there is another location available */
                            creator
                     FROM   encounter
                     WHERE  FIND_IN_SET(encounter_type, @encounter_type_exclusions) = 0
                     GROUP  BY patient_id,
                               Date(encounter_datetime)
+                              -- TODO: Group by location as well 
                 ) AS e;
             
             -- Add visit IDs to their encounters
--- a/etl/src/main/java/org/pih/hivmigration/etl/sql/VisitMigrator.groovy
+++ b/etl/src/main/java/org/pih/hivmigration/etl/sql/VisitMigrator.groovy
@@ -1,0 +1,42 @@
+package org.pih.hivmigration.etl.sql
+
+class VisitMigrator extends SqlMigrator {
+
+    void migrate() {
+
+        executeMysql("Create visits from existing encounters", '''
+            -- Get visit type "Clinic or Hospital Visit"
+            SET @visit_type_id = (SELECT visit_type_id FROM visit_type WHERE uuid = 'f01c54cb-2225-471a-9cd5-d348552c337c');
+            SET @encounter_type_exclusions = '';  -- a comma-separated string like '1,2,3'
+            
+            INSERT INTO visit
+            (patient_id,   visit_type_id,  date_started,   date_stopped,   location_id,   creator,   date_created, voided, uuid)
+            SELECT  e.patient_id, @visit_type_id, e.date_started, e.date_stopped, e.location_id, e.creator, now(),     0, uuid()
+            FROM
+                (
+                    SELECT patient_id,
+                           Subtime(Min(encounter_datetime), '00:05:00') AS date_started,  /* visit must start before first enc */
+                           Addtime(Max(encounter_datetime), '00:05:00') AS date_stopped,  /* visit must end after last enc */
+                           Max(location_id) as location_id,  /* Avoid using 'Unknown Location' if there is another location available */
+                           creator
+                    FROM   encounter
+                    WHERE  FIND_IN_SET(encounter_type, @encounter_type_exclusions) = 0
+                    GROUP  BY patient_id,
+                              Date(encounter_datetime)
+                ) AS e;
+            
+            -- Add visit IDs to their encounters
+            UPDATE encounter e
+                INNER JOIN visit v
+                ON e.patient_id = v.patient_id
+                    AND Date(e.encounter_datetime) = Date(v.date_started)
+            SET    e.visit_id = v.visit_id
+            WHERE FIND_IN_SET(e.encounter_type, @encounter_type_exclusions) = 0;
+        ''')
+
+    }
+
+    void revert() {
+        clearTable("visit")
+    }
+}
--- a/etl/src/main/java/org/pih/hivmigration/etl/sql/VisitMigrator.groovy
+++ b/etl/src/main/java/org/pih/hivmigration/etl/sql/VisitMigrator.groovy
@@ -15,8 +15,8 @@ class VisitMigrator extends SqlMigrator {
             FROM
                 (
                     SELECT patient_id,
-                           Subtime(Min(encounter_datetime), '00:05:00') AS date_started,  /* visit must start before first enc */
-                           Addtime(Max(encounter_datetime), '00:05:00') AS date_stopped,  /* visit must end after last enc */
+                           Min(encounter_datetime) AS date_started,  -- encounter_datetime is always midnight
+                           Addtime(Max(encounter_datetime), '23:59:00') AS date_stopped,
                            Max(location_id) as location_id,  /* Avoid using 'Unknown Location' if there is another location available */
                            creator
                     FROM   encounter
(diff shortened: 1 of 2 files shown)


etl/src/main/java/org/pih/hivmigration/etl/sql/VisitMigrator.groovy
```diff
@@ -0,0 +1,42 @@
...
+            SET @encounter_type_exclusions = '';  -- a comma-separated string like '1,2,3'
+            
+            INSERT INTO visit
+            (patient_id,   visit_type_id,  date_started,   date_stopped,   location_id,   creator,   date_created, voided, uuid)
+            SELECT  e.patient_id, @visit_type_id, e.date_started, e.date_stopped, e.location_id, e.creator, now(),     0, uuid()
+            FROM
+                (
+                    SELECT patient_id,
+                           Subtime(Min(encounter_datetime), '00:05:00') AS date_started,  /* visit must start before first enc */
+                           Addtime(Max(encounter_datetime), '00:05:00') AS date_stopped,  /* visit must end after last enc */
```
mogoodrich, line 19: Per our discussion today, based on how we are doing retrospective visits in the PIH EMR, the start date should have a time component of "00:00:00" and the end date should have one of "23:59:59"

etl/src/main/java/org/pih/hivmigration/etl/sql/VisitMigrator.groovy
```diff
@@ -0,0 +1,42 @@
+package org.pih.hivmigration.etl.sql
+
+class VisitMigrator extends SqlMigrator {
```
mseaton, line 3: At a high-level, one question I have is whether we want to do this as a one-off migration, or if this kind of thing can be something that just runs in OpenMRS.  Eg. on a schedule, find all encounters of a type that expect association with a visit and if they are not already associated with a visit, find a matching visit on the same date and location to associate with, and if one is not found, create one and associate it with it.  Have this just always running in the background.

etl/src/main/java/org/pih/hivmigration/etl/sql/VisitMigrator.groovy
```diff
@@ -0,0 +1,42 @@
...
+                    WHERE  FIND_IN_SET(encounter_type, @encounter_type_exclusions) = 0
+                    GROUP  BY patient_id,
+                              Date(encounter_datetime)
+                ) AS e;
+            
+            -- Add visit IDs to their encounters
+            UPDATE encounter e
+                INNER JOIN visit v
+                ON e.patient_id = v.patient_id
+                    AND Date(e.encounter_datetime) = Date(v.date_started)
```
mseaton, line 32: Likely we also want to match on location here?

etl/src/main/java/org/pih/hivmigration/etl/sql/VisitMigrator.groovy
```diff
@@ -0,0 +1,42 @@
...
+            SET    e.visit_id = v.visit_id
+            WHERE FIND_IN_SET(e.encounter_type, @encounter_type_exclusions) = 0;
```
mseaton, line 34: I don't think this extra check on encounter type here is needed.

etl/src/main/java/org/pih/hivmigration/etl/sql/VisitMigrator.groovy
```diff
@@ -0,0 +1,42 @@
...
+            SET @encounter_type_exclusions = '';  -- a comma-separated string like '1,2,3'
+            
+            INSERT INTO visit
+            (patient_id,   visit_type_id,  date_started,   date_stopped,   location_id,   creator,   date_created, voided, uuid)
+            SELECT  e.patient_id, @visit_type_id, e.date_started, e.date_stopped, e.location_id, e.creator, now(),     0, uuid()
+            FROM
+                (
+                    SELECT patient_id,
+                           Min(encounter_datetime) AS date_started,  -- encounter_datetime is always midnight
+                           Addtime(Max(encounter_datetime), '23:59:00') AS date_stopped,
```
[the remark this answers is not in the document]
mogoodrich: Yes, @mseaton  is correct... important because it makes it consistent with all the existing retro visits in the PIH EMR.

etl/src/main/java/org/pih/hivmigration/etl/sql/VisitMigrator.groovy
mogoodrich: Yeah, this is some example code that Brandon found on the OpenMRS wiki... doesn't do anything now, but would if we set up the encounter type exclusions above.

etl/src/main/java/org/pih/hivmigration/etl/sql/VisitMigrator.groovy
mogoodrich: It's an interesting point.  If I'm following the "insert" above does not group by location, but if there are multiple encounter locations within the same day it just picks one arbitrarily (or, more exactly, the one with the highest primary key, to make sure that "Uknown Location" is never chosen if there's another option).  So therefore we don't want to match on location here.

Assumedly, we shouldn't have multiple locations for the same patient on the same day, but @brandones probably worth running a quick query to see if this is the case?

etl/src/main/java/org/pih/hivmigration/etl/sql/VisitMigrator.groovy
mogoodrich: Yeah, per our discussion yesterday I think improving our logic to link encounters to visits would be valuable and something we should strongly consider. But I think it's tangential to this PR because I'd likely be in favor of building some sort of scheduled task or other functionality in OpenMRS rather than running the Migrator code at regular intervals.

mseaton: @mogoodrich I agree.  My point was more - if we have a scheduled task that will automatically link encounters to new or existing visits, we don't need this migration at all.

etl/src/main/java/org/pih/hivmigration/etl/sql/VisitMigrator.groovy
mseaton: Fair enough that there shouldn't be a visit at more than one location on the same day, and would be good to check.  But regardless, adding the group by on location above then wouldn't hurt, would it?  And then adding that in here too.

etl/src/main/java/org/pih/hivmigration/etl/sql/VisitMigrator.groovy
mseaton: My point was that we don't need this here since we are already doing it earlier above when we create visits.  By adding it here I think it is not needed because a) since we are migrating into an empty DB, the only visits that will exist are the one's we've already created, and we don't need to restrict further, and b) presumably we want to link any encounter to an existing visit if it's patient, date, and location match, even if it isn't in these encounter types...

mogoodrich: I think the idea here is that there may be some encounter types that never should be associated with visits... which is likely irrelevant in our case, but comes from the example code.

etl/src/main/java/org/pih/hivmigration/etl/sql/VisitMigrator.groovy
mogoodrich: Hmm... this is making me go down a bit of a rabbit hole re: visit locations... will bring it up after standup today.

etl/src/main/java/org/pih/hivmigration/etl/sql/VisitMigrator.groovy
mseaton: I think there are encounter types that should not, on their own, create a visit or imply the patient came that day.  But if a visit exists already that matches the encounter date and patient and location, we could consider associating  it with it.  Not a big deal I guess.

etl/src/main/java/org/pih/hivmigration/etl/sql/VisitMigrator.groovy
mogoodrich: Okay, scratch my previous comment... yes, I think we should try grouping by location, we just need to figure out how to handle any encounters at the Unknown Location, or any locations that aren't Visit locations.  Maybe handle that as a separate ticket if it's blocking this one too much.  @brandones  can you run quick distinct query count of all the encouter_locations in the system?

mogoodrich: Actually, looks like I already did the analysis of multiple encounters on the same day at different locations as part of this ticket:
https://pihemr.atlassian.net/browse/UHM-3239

mogoodrich (approved):
I think we are good to merge this in and then handle the location issues in follow-on tickets?

mogoodrich:
Strange @brandones  I just got an email saying you are requesting a PR on this one?  Just something that came through (very) late or did you mean to request a review on something else?

PR #18, opened 2020-09-03
UHM-4758 Add de-identification step to the migration pipeline

mogoodrich:


etl/src/main/java/org/pih/hivmigration/etl/sql/DeIdentifyMigrator.groovy
```diff
@@ -0,0 +1,100 @@
...
+                    WHEN p.source_patient_id % 20 = 12 THEN 'Alejandro'
+                    WHEN p.source_patient_id % 20 = 13 THEN 'Hugo'
+                    WHEN p.source_patient_id % 20 = 14 THEN 'Pablo'
+                    WHEN p.source_patient_id % 20 = 15 THEN 'Martin'
+                    WHEN p.source_patient_id % 20 = 16 THEN 'Innocent'
+                    WHEN p.source_patient_id % 20 = 17 THEN 'Alvaro'
+                    WHEN p.source_patient_id % 20 = 18 THEN 'Lucas'
+                    WHEN p.source_patient_id % 20 = 19 THEN 'Jamie'
+                END)
+            where given_name != 'UNKNOWN' and pn.person_id = p.person_id and p.gender != 'F'
```
mogoodrich, line 88: specifically did != 'F' so that 'U" is captured

etl/src/main/java/org/pih/hivmigration/etl/sql/Migrator.java
```diff
@@ -83,13 +86,17 @@ public void main() {
                 revert(new UserMigrator());
             }
             if (!shouldRevertOnly) {
-                migrate(new UserMigrator(), limit);
+                migrate(new UserMigrator(), -1);
```
mogoodrich, line 89: Unrelated to de-identification, I changed User and StagingData to ignore the limit, since they are quick and you can run into key errors if  users are missing.

etl/src/main/java/org/pih/hivmigration/etl/sql/DeIdentifyMigrator.groovy
```diff
@@ -0,0 +1,98 @@
...
+                    WHEN p.source_patient_id % 20 = 18 THEN 'Brady'
+                    WHEN p.source_patient_id % 20 = 19 THEN 'Lannister'
+                END)
+            where family_name != 'UNKNOWN' and pn.person_id = p.person_id
+            and pn.person_id not in (select person_id from users);
+            
+            update person_name pn, hivmigration_patients p
+            set given_name =
+                (CASE
+                    WHEN p.source_patient_id % 19 = 0 THEN 'Lucia'
```
mogoodrich, line 40: Do this modulo 19 instead of 20 so that family/given names don't always match

brandones (approved):

mogoodrich:
Thanks @brandones , going to go ahead and merge this in.

PR #19, opened 2020-09-04
UHM-4797 Migrate "Migrate Lab Results" to SQL Migrator

brandones:
https://pihemr.atlassian.net/browse/UHM-4797

This code is liable to change when I talk to Ellen about the new data model. However, it's likely to be structured much like this. This is the logic that was in Pentaho. All of the concepts referenced are already present.


etl/src/main/java/org/pih/hivmigration/etl/sql/LabResultMigrator.groovy
```diff
@@ -0,0 +1,261 @@
...
+            -- Fill in the constructs
+            --
+            -- Now for each construct we create the expected members.
+            
+            SET @specimen_number = (SELECT concept_id FROM concept WHERE uuid='162086AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA');
+            
+            -- Viral Load
+            --
+            
+            -- Add specimen number
```
mseaton, line 127: It seems unnecessary to repeat this population of specimen number for each test type individually.  This same query is repeated for each.  Why do that?  Furthermore, I suspect that this will result in the creation of obs with null values, since this doesn't do any checking that specimen number is actually populated.

I'd remove all of the individual queries, and just have a single section for "migrate specimen numbers", which migrates everything from hivmigration_lab_results where sample_id is not null.

etl/src/main/java/org/pih/hivmigration/etl/sql/LabResultMigrator.groovy
```diff
@@ -0,0 +1,261 @@
...
+            SET @hvl_value = (SELECT concept_id FROM concept WHERE uuid='3cd4a882-26fe-102b-80cb-0017a47871b2');
+            INSERT INTO obs(person_id, concept_id, encounter_id, obs_datetime, location_id, obs_group_id, value_numeric, creator, date_created, voided, uuid)
+            SELECT p.person_id, @hvl_value, e.encounter_id, r.obs_datetime, e.location_id, r.obs_id, r.value_numeric,1, now(), 0, uuid()
+            FROM hivmigration_lab_results r
+                     JOIN hivmigration_patients p ON r.source_patient_id = p.source_patient_id
+                     JOIN hivmigration_encounters e ON r.source_encounter_id = e.source_encounter_id
+            WHERE r.value_numeric IS NOT NULL AND r.test_type ='viral_load';
+            
+            -- Add undetectable VL values
+            SET @hvl_qualitative = (SELECT concept_id FROM concept WHERE uuid='1305AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA');
```
mseaton, line 145: One question I'd ask - and I'd like to see how we do this in Malawi @cioan / @mogoodrich / @dmdesimone / @lnball .  If a viral load result is detectable, and we store the numeric value as a viral load obs, do we also store the qualitative result as "DETECTABLE" in addition to that?  Or do we only capture this when the answer is NOT_DETECTED?

etl/src/main/java/org/pih/hivmigration/etl/sql/LabResultMigrator.groovy
```diff
@@ -0,0 +1,261 @@
...
+                     JOIN hivmigration_patients p ON r.source_patient_id = p.source_patient_id
+                     JOIN hivmigration_encounters e ON r.source_encounter_id = e.source_encounter_id
+            WHERE r.vl_detectable_lower_limit IS NOT NULL AND r.test_type ='viral_load';
+            
+            
+            -- CD4 Count
+            --
+            
+            -- Add specimen number
+            INSERT INTO obs(person_id, concept_id, encounter_id, obs_datetime, location_id, obs_group_id, value_text, creator, date_created, voided, uuid)
```
mseaton, line 168: See comment above about specimen number

etl/src/main/java/org/pih/hivmigration/etl/sql/LabResultMigrator.groovy
```diff
@@ -0,0 +1,261 @@
...
+                     JOIN hivmigration_patients p ON r.source_patient_id = p.source_patient_id
+                     JOIN hivmigration_encounters e ON r.source_encounter_id = e.source_encounter_id
+            WHERE r.value_numeric IS NOT NULL AND r.test_type ='cd4';
+            
+            
+            -- Hematocrit
+            --
+            
+            -- Specimen Number
+            INSERT INTO obs(person_id, concept_id, encounter_id, obs_datetime, location_id, obs_group_id, value_text, creator, date_created, voided, uuid)
```
mseaton, line 189: See comment above about specimen number

etl/src/main/java/org/pih/hivmigration/etl/sql/LabResultMigrator.groovy
```diff
@@ -0,0 +1,261 @@
...
+                     JOIN hivmigration_patients p ON r.source_patient_id = p.source_patient_id
+                     JOIN hivmigration_encounters e ON r.source_encounter_id = e.source_encounter_id
+            WHERE r.value_numeric IS NOT NULL AND r.test_type ='hematocrit';
+            
+            
+            -- PPD
+            --
+            
+            -- Specimen number
+            INSERT INTO obs(person_id, concept_id, encounter_id, obs_datetime, location_id, obs_group_id, value_text, creator, date_created, voided, uuid)
```
mseaton, line 210: See comment above about specimen number

etl/src/main/java/org/pih/hivmigration/etl/sql/LabResultMigrator.groovy
```diff
@@ -0,0 +1,261 @@
...
+                     JOIN hivmigration_patients p ON r.source_patient_id = p.source_patient_id
+                     JOIN hivmigration_encounters e ON r.source_encounter_id = e.source_encounter_id
+            WHERE r.value_numeric IS NOT NULL AND r.test_type ='ppd';
+            
+            
+            -- Rapid Test
+            --
+            
+            -- Specimen number
+            INSERT INTO obs(person_id, concept_id, encounter_id, obs_datetime, location_id, obs_group_id, value_text, creator, date_created, voided, uuid)
```
mseaton, line 232: See comment above about specimen number

etl/src/main/java/org/pih/hivmigration/etl/sql/LabResultMigrator.groovy
```diff
@@ -0,0 +1,261 @@
...
+                        ELSE @negative
+                       END AS value_coded
+                    ,1, now(), 0, uuid()
+            FROM hivmigration_lab_results r
+                JOIN hivmigration_patients p ON r.source_patient_id = p.source_patient_id
+                JOIN hivmigration_encounters e ON r.source_encounter_id = e.source_encounter_id
+            WHERE r.test_type ='tr';
+        ''')
+    }
+
```
mseaton, line 255: My biggest overall comment from reviewing the above, is that we desperately need some well tested functions that will allow us to migrate obs consistently and repeatedly and which will handle all of the boilerplate that you are copying from query to query.  This will:
a) allow us to have much less buggy code, as we will be writing methods once and testing them without worrying about whether every detail is correct each time
b) allow us to make changes to how we want to handle obs migrations more easily, but just changing the core functions that everything uses, rather than have to change every obs migration statement
c) allow the code to be infinitely more readable and easier to review the meaningful bits - eg. what concepts are we migrating into, what data are we migrating from, etc.

Can you please create some common functions for creating numeric, coded, boolean (which would convert a value of 1 or 't' into the true concept, etc), date, and text obs - and then use those functions here?

mseaton (changes requested):
Thanks @brandones .  See review comments below...

etl/src/main/java/org/pih/hivmigration/etl/sql/LabResultMigrator.groovy
mseaton: And by functions, I mean functions or procedures - I assume procedures is more appropriate.

etl/src/main/java/org/pih/hivmigration/etl/sql/LabResultMigrator.groovy
cioan: In Malawi EMR, we do not capture "DETECTABLE":
https://github.com/PIH/openmrs-module-pihmalawi/blob/master/omod/src/main/webapp/resources/htmlforms/viral_load_test_results.xml#L151

https://bwenzi.pih-emr.org/openmrs/dictionary/concept.htm?conceptId=8628

etl/src/main/java/org/pih/hivmigration/etl/sql/LabResultMigrator.groovy
```diff
@@ -0,0 +1,261 @@
...
+            order by d.patient_id desc, e.encounter_id desc
+        ''')
+
+        executeMysql("Add UUIDs", "UPDATE hivmigration_lab_results SET uuid = uuid();")
+
+        executeMysql("Load viral load results as observations", '''
+            --
+            -- Create construct obs
+            --
+            -- These have to be created first, since their obs_ids are fixed. The obs group elements for
```
mogoodrich, line 69: When you say obs ids are fixed, you mean that they need to link back to the keys created in the hivmigration_labs_result, correct?

etl/src/main/java/org/pih/hivmigration/etl/sql/LabResultMigrator.groovy
```diff
@@ -0,0 +1,261 @@
...
+            -- each construct are created later, and will receive sequential autogenerated obs_ids.
```
mogoodrich, line 70: It's worth noting that the lab workflow module, and the existing Lab Results form, does not use constructs to "wrap" results.  We will have to support constructs of some kind or another for tests that have multiple data points, but it might make sense to keep it simple where possible (ie no construct for single answers).

etl/src/main/java/org/pih/hivmigration/etl/sql/LabResultMigrator.groovy
mogoodrich: Following on my comment about constructs above, in the Lab Workflow module the specimen number would be a top-level property of the encounter, not of the individual tests, and this would allow us to avoid creating constructs for all these results.

Tangent: we should have a Confluence page (if we don't already) about our lab results modellign strategy.

etl/src/main/java/org/pih/hivmigration/etl/sql/LabResultMigrator.groovy
```diff
@@ -0,0 +1,261 @@
...
+                obs_datetime,
+                value_numeric,
+                value_text,
+                value_boolean,
+                vl_beyond_detectable_limit,
+                vl_detectable_lower_limit)
+            values (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
+        ''', '''
+            select   e.patient_id as source_patient_id,
+                     e.ENCOUNTER_ID as source_encounter_id,
```
mogoodrich, line 42: Assumedly the source encounter ID is either type "lab_result" or "anlap_lab_result", which we are already migrating to the Lab Results encounter type in the encounter migrator?  We shoudl just double-check this, to make sure we don't get any obs that are floating without encounters.  (Which might be okay, but if we aren't aware of it and plan for it, they will likely escape into the ether)

mogoodrich (commented):
Generally looks good... left some questions/comments.

Seems like the goal of this ticket/PR should be to get it to the point that migrated data ends up being viewing in and editable from the Lab Results form in OpenMRS.

etl/src/main/java/org/pih/hivmigration/etl/sql/LabResultMigrator.groovy
brandones: @mogoodrich yeah, agreed.  Ellen also isn't aware of any reason we'd need the constructs with specimen numbers. @mseaton any reason I shouldn't get rid of them?

mseaton: @brandones - are you implying that we should just discard the sample ids and not migrate them at all?

brandones: @mseaton  No, I'm saying that based on what I've heard from Mark and Ellen, that we probably don't need to create these constructs with sample ids migrated as specimen numbers. We can do anything we like with the sample ids. I think Mark's suggestion sounds reasonable.

mseaton: I see @brandones , yeah that makes sense.  Sorry, I didn't see @mogoodrich comments.

etl/src/main/java/org/pih/hivmigration/etl/sql/LabResultMigrator.groovy
brandones: I created a procedure. I'm not sure, but it might be less general than you were imagining—it's really just for `hivmigration_lab_results`. We could make it very general by abstracting out that table name, but I think this abstraction is already near the line of being more complicated than it's worth.

etl/src/main/java/org/pih/hivmigration/etl/sql/LabResultMigrator.groovy
brandones: Yes

brandones:
Ok @mogoodrich @mseaton I've rewritten this in terms of invocations of a procedure. The procedure is pretty ugly, since it abstracts out column names. I'm fine with it either way—with the procedure or no. Let me know what you think.

etl/src/main/java/org/pih/hivmigration/etl/sql/LabResultMigrator.groovy
mseaton: Yeah, I realized about 5 minutes after I made this comment that this was difficult to make generic at the bulk level, and I don't think what you introduced is really improving things.  If anything, I think this makes things more complicated.  Let's brainstorm a little together later.

etl/src/main/java/org/pih/hivmigration/etl/sql/LabResultMigrator.groovy
```diff
@@ -0,0 +1,193 @@
+package org.pih.hivmigration.etl.sql
+
+class LabResultMigrator extends SqlMigrator {
+
+    @Override
+    def void migrate() {
+        executeMysql("Create staging table", '''
+            create table hivmigration_lab_results (
+              obs_id int PRIMARY KEY AUTO_INCREMENT,
```
mseaton, line 9: Note, each row in this table often represents an obs group with multiple obs within it.  I don't know if we need to explicitly set obs ids for anything, and can let them autogenerate.  We can keep uuids for all of the possible obs that each row might create, and this will allow us to set these and also use them to associate obs and obs groups together.

etl/src/main/java/org/pih/hivmigration/etl/sql/LabResultMigrator.groovy
```diff
@@ -0,0 +1,193 @@
...
+              test_type VARCHAR(16), -- viral_load, CD4, tr, ppd, hematocrit
+              obs_datetime date,
+              value_numeric DOUBLE,
+              value_text VARCHAR(100),
+              value_boolean BOOLEAN,
+              vl_beyond_detectable_limit BOOLEAN,
+              vl_detectable_lower_limit DOUBLE,
+              tmp_vl_beyond_detectable_limit_concept int,
+              tmp_rapid_test_value int,
+              uuid char(38),
```
mseaton, line 23: Same comment here with uuid as my comment above with obs_id.  What exactly does this uuid represent?  We should be explicit, and possible have multiple obs fields (some of which might be null).  Eg. sample_id_uuid, quantitative_result uuid, qualitative_result_uuid, etc.

etl/src/main/java/org/pih/hivmigration/etl/sql/LabResultMigrator.groovy
```diff
@@ -0,0 +1,193 @@
...
+              sample_id VARCHAR(20),
+              test_type VARCHAR(16), -- viral_load, CD4, tr, ppd, hematocrit
+              obs_datetime date,
+              value_numeric DOUBLE,
+              value_text VARCHAR(100),
+              value_boolean BOOLEAN,
+              vl_beyond_detectable_limit BOOLEAN,
+              vl_detectable_lower_limit DOUBLE,
+              tmp_vl_beyond_detectable_limit_concept int,
+              tmp_rapid_test_value int,
```
mseaton, line 22: I would think that these two tmp_* columns are not needed any more, if we go with the Stored Proc approach?

etl/src/main/java/org/pih/hivmigration/etl/sql/LabResultMigrator.groovy
```diff
@@ -0,0 +1,193 @@
...
+                        end as vl_beyond_detectable_limit,                  
+                     r.RESULT_UNDETECTABLE_RANGE as vl_detectable_lower_limit
+            from     hiv_demographics_real d, hiv_encounters e, hiv_lab_results r, hiv_lab_tests t
+            where    d.patient_id = e.patient_id
+            and      e.encounter_id = r.encounter_id
+            and      r.lab_test_id = t.lab_test_id  
+            order by d.patient_id desc, e.encounter_id desc
+        ''')
+
+        executeMysql("Add UUIDs", "UPDATE hivmigration_lab_results SET uuid = uuid();")
```
mseaton, line 65: As above, might need to look at uuid population

etl/src/main/java/org/pih/hivmigration/etl/sql/LabResultMigrator.groovy
```diff
@@ -0,0 +1,193 @@
...
+                    a.concept_id, o.value_numeric, o.value_datetime, o.value_text, 1, e.date_created, 0, uuid()
+                FROM tmp_obs o
+                    JOIN       hivmigration_patients p ON o.source_patient_id = p.source_patient_id
+                    JOIN       hivmigration_encounters e ON o.source_encounter_id = e.source_encounter_id
+                    LEFT JOIN  concept q ON q.uuid = o.concept_uuid
+                    LEFT JOIN  concept a ON a.uuid = o.value_coded_uuid;
+            END $$
+            DELIMITER ;
+        ''')
+
```
mseaton, line 103: I would expect the above 2 stored procedures to be defined and executed outside of this particular job, so that they can just be available and used by any of the individual jobs that need them.  The behavior should be consistent whether we are migrating viral loads, med pickups, or any other obs.

etl/src/main/java/org/pih/hivmigration/etl/sql/LabResultMigrator.groovy
```diff
@@ -0,0 +1,193 @@
...
+            --
+            INSERT INTO tmp_obs (value_numeric, source_patient_id, source_encounter_id, concept_uuid)
+            SELECT value_numeric, source_patient_id, source_encounter_id, '3ceda710-26fe-102b-80cb-0017a47871b2'
+            FROM hivmigration_lab_results
+            WHERE test_type = 'cd4'
+            
+            -- etc. for the others
+            
+            CALL migrate_tmp_obs();
+
```
mseaton, line 156: Yes, this generally looks right above.

mseaton (changes requested):

etl/src/main/java/org/pih/hivmigration/etl/sql/LabResultMigrator.groovy
mseaton: OK, looks like you handle this below by using this obs_id as the group obs_id.  Might want to just rename this column to "obs_group_id" to be explicit about it.

etl/src/main/java/org/pih/hivmigration/etl/sql/LabResultMigrator.groovy
```diff
@@ -0,0 +1,193 @@
...
+            END $$
+            CREATE PROCEDURE migrate_tmp_obs()
+            BEGIN
+                INSERT INTO obs (
+                    obs_id, person_id, encounter_id, obs_group_id, obs_datetime, location_id, concept_id,
+                    value_coded, value_numeric, value_datetime, value_text, creator, date_created, voided, uuid
+                )
+                SELECT
+                    o.obs_id, p.person_id, e.encounter_id, o.obs_group_id, e.encounter_date, e.location_id, q.concept_id,
+                    a.concept_id, o.value_numeric, o.value_datetime, o.value_text, 1, e.date_created, 0, uuid()
```
mseaton, line 94: My guess is that the "uuid()" function here probably should be replaced with the uuid that is being populated on the hivmigration_lab_results table, and that this represents the uuid of the group.  Correct?

etl/src/main/java/org/pih/hivmigration/etl/sql/LabResultMigrator.groovy
brandones: Oh whoops, UUIDs are now added via the procedure, so this column isn't necessary at all any more.

etl/src/main/java/org/pih/hivmigration/etl/sql/LabResultMigrator.groovy
brandones: The obs group has an explicitly set `obs_id`, and its members have explicitly set `obs_group_id`s, which are that group's `obs_id`. It's exactly the way one creates obs normally, and a way of improving on it isn't obvious to me.

etl/src/main/java/org/pih/hivmigration/etl/sql/LabResultMigrator.groovy
brandones: Yeah, agreed, just getting this hacked together right now

etl/src/main/java/org/pih/hivmigration/etl/sql/LabResultMigrator.groovy
mseaton: Yeah, as I wrote above, maybe just renaming it to obs_group_id is more clearer

mseaton (approved):
LGTM, thanks @brandones !

brandones:
Thanks for all the feedback and direction!

PR #20, opened 2020-09-04
UHM-4828: Set Encounter Location on Encounters

mogoodrich:


etl/src/main/java/org/pih/hivmigration/etl/sql/StagingDataMigrator.groovy
```diff
@@ -9,107 +9,121 @@ class StagingDataMigrator extends SqlMigrator {
                 openmrs_id int NOT NULL
             );
             
+            # Unknown location
             insert into hivmigration_health_center(hiv_emr_id, openmrs_id)
```
mogoodrich, line 13: Switched all these to be uuid-based, to protect against changing names breaking (siliently) our migration.

mseaton (approved):
LGTM

brandones (approved):
Nice! Thanks!

PR #25, opened 2020-09-10
UHM-4799, migrate HIV_DISPENSING_MEDS to OpenMRS

cioan:
Hi @mseaton , this is just the first step that brings over to OpenMRS the HIV_DISPENSING and HIV_DISPENSING_MEDS. I am sure some column names might need to change? But, I hope everything else is according to what we discussed yesterday. Thanks. (the full import ran on my machine in 2:30)


mseaton (approved):
Perfect, thanks @cioan

PR #27, opened 2020-09-11
UHM-4800 Migrate "Migrate Vitals" to SQL Migrator

brandones:
https://pihemr.atlassian.net/browse/UHM-4800

### DIFF
--- a/etl/src/main/java/org/pih/hivmigration/etl/sql/VitalsMigrator.groovy
+++ b/etl/src/main/java/org/pih/hivmigration/etl/sql/VitalsMigrator.groovy
@@ -1,0 +1,85 @@
+package org.pih.hivmigration.etl.sql
+
+class VitalsMigrator extends SqlMigrator {
+    @Override
+    def void migrate() {
+
+        executeMysql("Create vitals staging table", '''
+            CREATE TABLE hivmigration_vitals (
+              source_encounter_id int,
+              sign varchar(30),
+              result double,
+              result_unit varchar(100),
+              source_patient_id int,
+              KEY `source_encounter_idx` (`source_encounter_id`)
+            );
+        ''')
+
+        // TODO: should we do something with the 563 vitals with encounter type 'food_study'
+        //      or the 3 with type patient_contact?
+        loadFromOracleToMySql('''
+            INSERT INTO hivmigration_vitals (source_encounter_id, sign, result, result_unit, source_patient_id)
+            VALUES (?, ?, ?, ?, ?)
+        ''', '''
+            select 
+                v.ENCOUNTER_ID, 
+                v.SIGN, 
+                v.RESULT, 
+                v.RESULT_UNIT,
+                e.PATIENT_ID
+            from HIV_EXAM_VITAL_SIGNS v, HIV_ENCOUNTERS e
+            where v.ENCOUNTER_ID = e.ENCOUNTER_ID 
+            and SIGN in ('height', 'weight', 'bmi')
+            and e.TYPE in ('intake', 'followup')
+            and e.PATIENT_ID is not null
+        ''')
+
+        executeMysql("CALL create_tmp_obs_table();")
+
+        setAutoIncrement("tmp_obs", "(SELECT max(obs_id) + 1 FROM obs)")
+
+        executeMysql("Add vitals obs", '''
+            INSERT INTO tmp_obs (source_patient_id, source_encounter_id, value_numeric, concept_uuid)
+            SELECT
+                source_patient_id,
+                source_encounter_id,
+                CASE
+                    -- result unit is unreliable, but fortunately the values are unambiguous
+                    WHEN result < 3.0 THEN round(result * 100, 1)
+                    WHEN result > 40 AND result < 300 THEN result
+                END,
+                '3ce93cf2-26fe-102b-80cb-0017a47871b2'
+            FROM hivmigration_vitals
+            WHERE sign = 'height'
+            AND (result < 3.0 OR (result > 40 AND result < 300));
+
+            INSERT INTO tmp_obs (source_patient_id, source_encounter_id, value_numeric, concept_uuid)
+            SELECT
+                source_patient_id,
+                source_encounter_id,
+                CASE
+                    WHEN result_unit = 'kgs' THEN result
+                    WHEN result_unit = 'lbs' THEN round(result * 0.453592, 1)
+                    -- Drop 3461 entries with result_unit = NULL
+                    -- result distribution for these has Avg 115, stddev 33.5. Not obvious what to use.
+                    -- (there are only 813 entries with result_unit not NULL)
+                END,
+                '3ce93b62-26fe-102b-80cb-0017a47871b2'
+            FROM hivmigration_vitals
+            WHERE sign = 'weight'
+            AND result_unit IN ('kgs', 'lbs');
+            
+            CALL migrate_tmp_obs();
+        ''')
+    }
+
+    @Override
+    def void revert() {
+        executeMysql('''
+            DELETE o FROM obs o JOIN concept c ON c.concept_id = o.concept_id
+            WHERE c.uuid IN ('3ce93cf2-26fe-102b-80cb-0017a47871b2', '3ce93b62-26fe-102b-80cb-0017a47871b2');
+        ''')
+        setAutoIncrement("obs", "(select (max(obs_id)+1) from obs)")
+        executeMysql("DROP TABLE IF EXISTS hivmigration_vitals;")
+    }
+}
--- a/etl/src/main/java/org/pih/hivmigration/etl/sql/VitalsMigrator.groovy
+++ b/etl/src/main/java/org/pih/hivmigration/etl/sql/VitalsMigrator.groovy
@@ -118,14 +118,11 @@ class VitalsMigrator extends SqlMigrator {
                 CASE
                     WHEN result_unit = 'kgs' THEN result
                     WHEN result_unit = 'lbs' THEN round(result * 0.453592, 1)
-                    -- Drop 3461 entries with result_unit = NULL
-                    -- result distribution for these has Avg 115, stddev 33.5. Not obvious what to use.
-                    -- (there are only 813 entries with result_unit not NULL)
+                    ELSE result  -- use kgs for the 3461 entries with result_unit = NULL
                 END,
                 '3ce93b62-26fe-102b-80cb-0017a47871b2'
             FROM hivmigration_vitals
-            WHERE sign = 'weight'
-            AND result_unit IN ('kgs', 'lbs');
+            WHERE sign = 'weight';
         ''')
 
         executeMysql("CALL migrate_tmp_obs_vitals();")
--- a/etl/src/main/java/org/pih/hivmigration/etl/sql/VitalsMigrator.groovy
+++ b/etl/src/main/java/org/pih/hivmigration/etl/sql/VitalsMigrator.groovy
@@ -150,9 +150,9 @@ class VitalsMigrator extends SqlMigrator {
                 source_patient_id,
                 source_encounter_id,
                 CASE
-                    WHEN result > 60 THEN  -- TODO
+                    WHEN result > 60 THEN round((result - 32.0) * 0.555555, 1)
                     ELSE result
-                END
+                END,
                 '3ce939d2-26fe-102b-80cb-0017a47871b2'
             FROM hivmigration_vitals
             WHERE sign = 'temperature';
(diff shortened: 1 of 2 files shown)


brandones:
@cioan See the revert logic here

etl/src/main/java/org/pih/hivmigration/etl/sql/VitalsMigrator.groovy
```diff
@@ -0,0 +1,85 @@
...
+            AND result_unit IN ('kgs', 'lbs');
+            
+            CALL migrate_tmp_obs();
+        ''')
+    }
+
+    @Override
+    def void revert() {
+        executeMysql('''
+            DELETE o FROM obs o JOIN concept c ON c.concept_id = o.concept_id
```
cioan, line 79: Thanks @brandones . This would work if this concept_id is used for obs with just this encounter type. In other situations, e.g. Next Visit Date concept id could be used for obs with different encounter types, so deleting all obs with a specific concept_id might not work in all situations.

brandones: Oh, that's true, and annoying

etl/src/main/java/org/pih/hivmigration/etl/sql/VitalsMigrator.groovy
```diff
@@ -0,0 +1,85 @@
...
+              sign varchar(30),
+              result double,
+              result_unit varchar(100),
+              source_patient_id int,
+              KEY `source_encounter_idx` (`source_encounter_id`)
+            );
+        ''')
+
+        // TODO: should we do something with the 563 vitals with encounter type 'food_study'
+        //      or the 3 with type patient_contact?
```
mseaton, line 19: I thought the decision you and @mogoodrich decided on was that these should result in Vitals encounters, not added to the Intake, Followup, or other encounters?  If so, then the source encounter type doesn't really matter, does it?  We should not be restricting based on this.

etl/src/main/java/org/pih/hivmigration/etl/sql/VitalsMigrator.groovy
```diff
@@ -0,0 +1,85 @@
...
+            select 
+                v.ENCOUNTER_ID, 
+                v.SIGN, 
+                v.RESULT, 
+                v.RESULT_UNIT,
+                e.PATIENT_ID
+            from HIV_EXAM_VITAL_SIGNS v, HIV_ENCOUNTERS e
+            where v.ENCOUNTER_ID = e.ENCOUNTER_ID 
+            and SIGN in ('height', 'weight', 'bmi')
+            and e.TYPE in ('intake', 'followup')
```
mseaton, line 33: Per my above comment, unless we are migrating these into intake and followup encounters, we should remove this restriction.

etl/src/main/java/org/pih/hivmigration/etl/sql/VitalsMigrator.groovy
```diff
@@ -0,0 +1,85 @@
...
+        ''', '''
+            select 
+                v.ENCOUNTER_ID, 
+                v.SIGN, 
+                v.RESULT, 
+                v.RESULT_UNIT,
+                e.PATIENT_ID
+            from HIV_EXAM_VITAL_SIGNS v, HIV_ENCOUNTERS e
+            where v.ENCOUNTER_ID = e.ENCOUNTER_ID 
+            and SIGN in ('height', 'weight', 'bmi')
```
mseaton, line 32: Can we just bang out this entire table migration now, instead of just migrating these 3 values?  There are not that many options, and I would expect we already have a clear migration for all of them.

select sign from HIV_EXAM_VITAL_SIGNS group by sign order by sign;

height,weight,bmi
+
blood_pressure_dias, blood_pressure_sys, heart_rate, respiration_rate, temperature

So we just need to add in migration for 5 additional concepts, which are all in our Vitals app in OpenMRS already.

Can we do this rather than leave this as a pending todo?  Especially since we are migrating to a Vitals encounter.

etl/src/main/java/org/pih/hivmigration/etl/sql/VitalsMigrator.groovy
```diff
@@ -0,0 +1,85 @@
...
+            and e.TYPE in ('intake', 'followup')
+            and e.PATIENT_ID is not null
+        ''')
+
```
mseaton, line 36: We'll want to first create new Vitals encounters here, or we should add explicitly into the EncounterMigrator to ensure it is already in place.

etl/src/main/java/org/pih/hivmigration/etl/sql/VitalsMigrator.groovy
```diff
@@ -0,0 +1,85 @@
...
+                source_encounter_id,
+                CASE
+                    -- result unit is unreliable, but fortunately the values are unambiguous
+                    WHEN result < 3.0 THEN round(result * 100, 1)
+                    WHEN result > 40 AND result < 300 THEN result
+                END,
+                '3ce93cf2-26fe-102b-80cb-0017a47871b2'
+            FROM hivmigration_vitals
+            WHERE sign = 'height'
+            AND (result < 3.0 OR (result > 40 AND result < 300));
```
mseaton, line 54: What is this all about?  So we're arbitrarily deciding not to migrate any data if the result is in the range of [3, 40] or >= 300?  Can we remove this?

etl/src/main/java/org/pih/hivmigration/etl/sql/VitalsMigrator.groovy
```diff
@@ -0,0 +1,85 @@
...
+
+        executeMysql("Add vitals obs", '''
+            INSERT INTO tmp_obs (source_patient_id, source_encounter_id, value_numeric, concept_uuid)
+            SELECT
+                source_patient_id,
+                source_encounter_id,
+                CASE
+                    -- result unit is unreliable, but fortunately the values are unambiguous
+                    WHEN result < 3.0 THEN round(result * 100, 1)
+                    WHEN result > 40 AND result < 300 THEN result
```
mseaton, line 49: This case statement needs to account for results outside of these ranges.  This is not covering all possible values.

etl/src/main/java/org/pih/hivmigration/etl/sql/VitalsMigrator.groovy
```diff
@@ -0,0 +1,85 @@
...
+            AND (result < 3.0 OR (result > 40 AND result < 300));
+
+            INSERT INTO tmp_obs (source_patient_id, source_encounter_id, value_numeric, concept_uuid)
+            SELECT
+                source_patient_id,
+                source_encounter_id,
+                CASE
+                    WHEN result_unit = 'kgs' THEN result
+                    WHEN result_unit = 'lbs' THEN round(result * 0.453592, 1)
+                    -- Drop 3461 entries with result_unit = NULL
```
mseaton, line 63: No, we're not dropping values.  We should migrate these in and highlight that these need to be looked at from a data quality perspective and validated.  It is likely that we can determine the correct units based on comparison with other data in the patient record.  What other weights did the patient have before/after?  Look at whether the value or converted value falls into the same range.  Or migrate as is in KG and leave this for a data cleaning exercise.  Don't discard the data.

etl/src/main/java/org/pih/hivmigration/etl/sql/VitalsMigrator.groovy
mseaton: Typically this is the point of populating a staging table with the uuid ahead of time and then inserting this uuid.  Then it's just a matter of deleting, joining on the staging table uuid.  This might be slow though.

mseaton (changes requested):
Lots of comments here @brandones .  Let me know if you want to discuss.  @mogoodrich  FYI

etl/src/main/java/org/pih/hivmigration/etl/sql/VitalsMigrator.groovy
mogoodrich: So, yes, assuming that we are sticking with our decision to store these in Vitals encounters, I'm thinking here that we have to create a new Vitals encounter for each source_encounter_id in the hivmigration_vitals table?  

Seems like we'd have to do this here, not in the EncounterMigrator.  If we did do this, we'd want the VisitMigrator to run after this (which I don't think should be an issue).

etl/src/main/java/org/pih/hivmigration/etl/sql/VitalsMigrator.groovy
mogoodrich: Sounds like @mseaton  will disagree but I'd be okay with dropping these for now if we provide a report of the bad values to the team on the ground and have a follow up ticket for deciding what we want to do about them.

etl/src/main/java/org/pih/hivmigration/etl/sql/VitalsMigrator.groovy
mogoodrich: @brandones @cioan not sure I follow here... we aren't restricting by encounter type here are we?  I guess the problem would come in if another *Migrator is migrating obs with the same concept?  That might be the case, but I don't think we've gotten there yet, so it might be just cross that bridge when we come to it?

mogoodrich (commented):
Thanks @brandones ...

Added my thoughts... generally agree with @mseaton  with the long-term solutions/what we can't skip, but I guess I'm more amendable to punting on certain things *if* we ticket what we are punting on.  Especially since "Vitals" is not a current priority, but this is more about getting away from the Pentaho tables.

Happy to discuss further...

Also, I'm in the midst of testing the Lab Results migration, so what I discover there may affect things... one thing I've discovered is that we need to set the location_id on the obs to avoid a stack trace when editing. (will put details on that ticket)

etl/src/main/java/org/pih/hivmigration/etl/sql/VitalsMigrator.groovy
mseaton: Unless these values are causing problems, my preference would be to include them...and then ticket as something to review and ensure are done properly, rather than to exclude them and then ticket.

etl/src/main/java/org/pih/hivmigration/etl/sql/VitalsMigrator.groovy
brandones: Sorry @mogoodrich , this conversation was in reference to some of @cioan 's work that he was talking about on Friday. This thread can be ignored.

etl/src/main/java/org/pih/hivmigration/etl/sql/VitalsMigrator.groovy
brandones: What do you think should be done with those values?

There are about 980 height values between 3 and 40. A handful are actually likely values for preemie neonates—so it's probably worthwhile to reduce 40 to 25.

There are 40 values above 300.

etl/src/main/java/org/pih/hivmigration/etl/sql/VitalsMigrator.groovy
brandones: Migrating as kg and ticketing seems reasonable to me. I wonder if we should figure out a way to make the pipeline produce a log file of data to review—something we could pass off to the team that's like

| patient | field | value |
|-------|----|----|
| 12345 | height | 472cm |
| 5423 | birthdate | 1800-01-01 |
...

mogoodrich: That's cool thought @brandones , having the pipeline produce a table/log of data to be validated.

etl/src/main/java/org/pih/hivmigration/etl/sql/VitalsMigrator.groovy
mseaton: Why do we need to do anything with them except migrate as-is?  Do they cause a problem in openmrs/

brandones: Well, there's a unit, but I guess we could just say "cm" for all the obviously invalid data. Seems fine.

brandones:
@mseaton @mogoodrich Ready for re-review. The vitals migrator takes about 20 minutes.

etl/src/main/java/org/pih/hivmigration/etl/sql/VitalsMigrator.groovy
```diff
@@ -0,0 +1,183 @@
...
+                uuid()
+            FROM hivmigration_vitals v
+                JOIN hivmigration_vitals_encounters ve ON v.source_encounter_id = ve.source_encounter_id
+                JOIN hivmigration_encounters e ON v.source_encounter_id = e.source_encounter_id
+                JOIN hivmigration_patients p ON e.source_patient_id = p.source_patient_id
+                JOIN hivmigration_users hu ON e.source_creator_id = hu.source_user_id
+            GROUP BY e.source_encounter_id;
+        ''')
+
+        executeMysql("Create migrate_tmp_obs variant for vitals migration", '''
```
mseaton, line 75: Can you explain why we have this stored procedure?  Since it is specialized for this job, and you only ever call it once, wouldn't it be easier and clearer to just execute this insert statement directly at the end of the job?

etl/src/main/java/org/pih/hivmigration/etl/sql/VitalsMigrator.groovy
```diff
@@ -0,0 +1,183 @@
...
+            WHERE sign = 'temperature';
+        ''')
+
+        executeMysql("CALL migrate_tmp_obs_vitals();")
+    }
+
+    @Override
+    def void revert() {
+        // This commented-out code takes a very long time. Use only if needed.
+//        executeMysql("Clear vitals obs", '''
```
mseaton, line 167: It's probably not faster, but I would not use the logic below.  I would just delete based on encounters.  You have the hivmigration_vitals_encounters that you created.  Just delete from obs where encounter_id is in hivmigration_vitals_encounters, and then delete from encounters where encounter_id in hivmigration_vitals_encounters.  Wouldn't that get everything?

mseaton (approved):
Approving, but with a few comments where I think there are some things you could improve.

etl/src/main/java/org/pih/hivmigration/etl/sql/VitalsMigrator.groovy
brandones: Ha, yep :facepalm:

etl/src/main/java/org/pih/hivmigration/etl/sql/VitalsMigrator.groovy
brandones: Sure, it ought to. But I don't think either solution is really acceptable, since they seem to cause mysql to hang. To actually get it to work we'd have to execute it in a loop with a limit `while` there are any matching rows left. I'm happy with `clearTable('obs')` for now.

brandones:
@mseaton I made one more change, fixing blood pressure values. I multiply them by 10 if they're below a certain threshold-- 20 for diastolic, 30 for systolic.

I'm going to go ahead and merge this, but if you think we should do something different with BP, I'm happy to fix it up after.

brandones:
Thanks for reviewing!

mogoodrich (commented):
Delayed review, but LGTM!

PR #28, opened 2020-09-11
UHM-4799, step 2 migrate dispensing obs

cioan:
Here is step#2 that creates all dispensing obs except the medication construct. I will work on this next. Thanks @mseaton , @mogoodrich and @brandones !


mseaton (approved):
LGTM

cioan:
Thanks @mseaton !

PR #29, opened 2020-09-14
UHM-4799

cioan:
I was able to insert the dispensing construct which involved inserting obs_group_id column.
@brandones , In order to be able to migrate obs from the tmp_obs table into the obs table, I had to disable the FOREIGN KEY check. I am wondering how this worked for the Lab Results?
I always got the following error message:

INFO    2020-09-12 10:49:23,540 [main] org.pih.hivmigration.etl.sql.SqlMigrator  - Executing: Load meds dispensing as observations
Exception in thread "main" java.sql.SQLException: Cannot add or update a child row: a foreign key constraint fails (`haitihivtest`.`obs`, CONSTRAINT `obs_grouping_id` FOREIGN KEY (`obs_group_id`) REFERENCES `obs` (`obs_id`)) Query: CALL migrate_tmp_obs() Parameters: []
	at org.apache.commons.dbutils.AbstractQueryRunner.rethrow(AbstractQueryRunner.java:392)
	at org.apache.commons.dbutils.QueryRunner.update(QueryRunner.java:491)
	at org.apache.commons.dbutils.QueryRunner.update(QueryRunner.java:377)
	at org.pih.hivmigration.etl.sql.SqlMigrator.executeMysql(SqlMigrator.java:103)
	at org.pih.hivmigration.etl.sql.SqlMigrator.executeMysql(SqlMigrator.java:112)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.codehaus.groovy.runtime.callsite.PlainObjectMetaMethodSite.doInvoke(PlainObjectMetaMethodSite.java:43)
	at org.codehaus.groovy.runtime.callsite.PogoMetaMethodSite$PogoCachedMethodSiteNoUnwrapNoCoerce.invoke(PogoMetaMethodSite.java:193)
	at org.codehaus.groovy.runtime.callsite.PogoMetaMethodSite.callCurrent(PogoMetaMethodSite.java:61)
Disconnected from the target VM, address: '127.0.0.1:63963', transport: 'socket'
	at org.codehaus.groovy.runtime.callsite.CallSiteArray.defaultCallCurrent(CallSiteArray.java:51)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.callCurrent(AbstractCallSite.java:171)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.callCurrent(AbstractCallSite.java:194)
	at org.pih.hivmigration.etl.sql.MedpickupsMigrator.migrate(MedpickupsMigrator.groovy:87)
	at org.pih.hivmigration.etl.sql.Migrator.migrate(Migrator.java:123)
	at org.pih.hivmigration.etl.sql.Migrator.main(Migrator.java:78)
	at org.pih.hivmigration.etl.sql.Migrator.main(Migrator.java:52)

Process finished with exit code 1

### DIFF
--- a/etl/src/main/java/org/pih/hivmigration/etl/sql/MedpickupsMigrator.groovy
+++ b/etl/src/main/java/org/pih/hivmigration/etl/sql/MedpickupsMigrator.groovy
@@ -186,5 +186,6 @@ class MedpickupsMigrator extends SqlMigrator{
     def void revert() {
         executeMysql("drop table if exists hivmigration_dispensing_meds")
         executeMysql("drop table if exists hivmigration_dispensing")
+        clearTable("obs")
     }
 }
--- a/etl/src/main/java/org/pih/hivmigration/etl/sql/MedpickupsMigrator.groovy
+++ b/etl/src/main/java/org/pih/hivmigration/etl/sql/MedpickupsMigrator.groovy
@@ -5,6 +5,53 @@ class MedpickupsMigrator extends SqlMigrator{
 
     @Override
     def void migrate() {
+
+        executeMysql("Create table for mapping HIV meds to OpenMRS Drugs", '''
+            create table hivmigration_openmrs_drugs (                            
+              hiv_med_name VARCHAR(100) PRIMARY KEY,
+              openmrs_drug_name VARCHAR(256),
+              openmrs_drug_uuid CHAR(38),
+              concept_uuid CHAR(38)              
+            );
+        ''')
+
+        executeMysql("Add meds mappings", '''
+            insert into hivmigration_openmrs_drugs(
+                hiv_med_name,
+                openmrs_drug_name,
+                openmrs_drug_uuid,
+                concept_uuid) 
+            values
+                ('3TC 10 mg sp','Lamivudine (3TC), Oral solution, 10mg/mL, 240mL bottle','78f96684-dfbe-11e9-8a34-2a2ae2dbcce4','3cd24e3e-26fe-102b-80cb-0017a47871b2'),
+                ('ABC 20 mg','Abacavir sulfate (ABC), Oral solution, 20mg/mL, 240mL bottle','78f98308-dfbe-11e9-8a34-2a2ae2dbcce4','70057AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA'),
+                ('ABC 300 mg','Abacavir sulfate (ABC), 300mg tablet','78f981d2-dfbe-11e9-8a34-2a2ae2dbcce4','70057AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA'),
+                ('ABC/3TC 60/30 mg','Abacavir sulfate (ABC) 60mg + Lamivudine (3TC) 30mg, tablet for oral suspension','78faa9b8-dfbe-11e9-8a34-2a2ae2dbcce4','103166AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA'),
+                ('ATV/r 300/100 mg','Atazanavir sulfate (ATV) 300mg + Ritonavir (r) 100mg tablet','78f95d38-dfbe-11e9-8a34-2a2ae2dbcce4','159809AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA'),
+                ('AZT 10 mg sp','Zidovudine (AZT), Oral solution, 10mg/mL, 240mL bottle','78f95bc6-dfbe-11e9-8a34-2a2ae2dbcce4','3cd444be-26fe-102b-80cb-0017a47871b2'),
+                ('AZT/3TC 300/150 mg','Lamivudine (3TC) 150mg + Zidovudine (AZT) 300mg tablet','78f968e6-dfbe-11e9-8a34-2a2ae2dbcce4','3cd25168-26fe-102b-80cb-0017a47871b2'),
+                ('AZT/3TC 60/30 mg','Lamivudine (3TC) 30mg + Zidovudine (AZT) 60mg tablet','78f95fa4-dfbe-11e9-8a34-2a2ae2dbcce4','3cd25168-26fe-102b-80cb-0017a47871b2'),
+                ('AZT/3TC/NVP 300/150/200 mg','Lamivudine (3TC) 150mg + Nevirapine (NVP) 200mg + Zidovudine (AZT) 300mg tablet','78f9739a-dfbe-11e9-8a34-2a2ae2dbcce4','3cdc4a42-26fe-102b-80cb-0017a47871b2'),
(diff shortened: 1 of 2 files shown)


etl/src/main/java/org/pih/hivmigration/etl/sql/ProcedureSetup.groovy
```diff
@@ -37,6 +38,7 @@ class ProcedureSetup extends SqlMigrator {
                     JOIN       hivmigration_encounters e ON o.source_encounter_id = e.source_encounter_id
                     LEFT JOIN  concept q ON q.uuid = o.concept_uuid
                     LEFT JOIN  concept a ON a.uuid = o.value_coded_uuid;
+                SET FOREIGN_KEY_CHECKS=1;
```
mseaton, line 41: @cioan - don't turn off foreign key checking.  This would be bad practice.  It would be better to figure out what the dependency is and resolve it.  In this case, my guess is that groups need to be inserted prior to obs the refer to them, right?  So can we just change this procedure to do that?  Either by having 2 different insert statements one after the other (the first inserts where obs_group_id is null, the second inserts where obs_group_id is not null), or (if it works) have a single insert statement that just orders the select query in this way?

mseaton (changes requested):
@cioan - see my comments. @brandones and @mogoodrich , thoughts?

etl/src/main/java/org/pih/hivmigration/etl/sql/MedpickupsMigrator.groovy
```diff
@@ -154,5 +186,6 @@ class MedpickupsMigrator extends SqlMigrator{
     def void revert() {
         executeMysql("drop table if exists hivmigration_dispensing_meds")
         executeMysql("drop table if exists hivmigration_dispensing")
+        clearTable("obs")
```
mogoodrich, line 189: Didn't we decide we wanted to try avoid clearing out all the obs if possible?  Or is that just too difficult to figure out how to do at this point? I could see the argument that this  becomes a bit of a rabbit hole.

I'm wondering if it would just be easier to create a two staging servers... one for migration, that gets wiped on each change to the migration pipeline, and another that *doesn't* have any migrated data it in but states stable and could be used for creating test data, etc... thoughts @dmdesimone  @cioan  @brandones  @mseaton ?

etl/src/main/java/org/pih/hivmigration/etl/sql/ProcedureSetup.groovy
mogoodrich: I think @cioan asked on another thread, but interested how @brandones handled it with Lab Results.

brandones: Like Mike says—the group needs to come before any obs that reference it.

etl/src/main/java/org/pih/hivmigration/etl/sql/MedpickupsMigrator.groovy
brandones: Not sure why you'd want to create non-ephemeral test data. The only things we need to be testing is whether migrated-in data looks right and whether new data can be entered on top of it. I'd like to keep "migrating data into a database that has other data in it" out of our scope of work.

It hasn't caused any problems for my development process to just clear the obs table, since so far we don't have different Migrators creating interdependent obs. And I don't think we should.

Feel free if you want to be more exact about what gets reverted in the obs table, but also feel free not to?

mogoodrich: Yeah, I'm fine with just clearing out the obs table but then creating another Haiti EMR Test instance if we need to... @brandones  for the non-ephemeral question, I guess it's not just creating long-standing test data, but if @dmdesimone  is testing a new form on Haiti HIV Test and at the same time someone commits a change to the pipeline and his test data gets blown away from underneath him while he's in the process of testing it causes confusion.

Another option would be to *only* run the migration nightly, not on each commit, but then it's harder (or at least there's a longer turnaround) to testing migration steps.

etl/src/main/java/org/pih/hivmigration/etl/sql/ProcedureSetup.groovy
cioan: Thanks @mseaton , @mogoodrich and @brandones ! Of course, it was a bug in the statement I use to create the dispensing construct. The tmp_obs table and procedure Brandon created works fine for dispensing meds obs too. No need to turn off FOREIGN KEY checks. Thanks!

cioan:
Hi @mseaton , @brandones and @mogoodrich , I think I have completed the Medication Orders construct and migrated all the HIV meds that already have Drugs in the OpenMRS table. I have used this temporary table to map the existing HIV meds to the OpenMRS drugs:
https://github.com/PIH/hivmigration/compare/UHM-4799#diff-d68e38771160bac769a7eb98862cd606R8

Once the following list of drugs is complete I will update the code(the temp table).
https://docs.google.com/spreadsheets/d/16cFAqy_mE-VHlMVLA9NOVrS4kCYzI8dK6rURzgNl7aA/edit?pli=1#gid=140586218

@brandones , I had to add another column, drug_uuid to the tmp_obs table to handle the obs with Drugs. Hope this does not break the lab results obs import. Thanks!

brandones:
@cioan Looks like it should be fine, thanks!

mogoodrich (approved):
LGTM, thanks @cioan !

It occurs to me that we won't be able to fully test this until we get the form fixed to support value drug, and adjust the way we do categories, but we will be doing that this sprint!

It occurs to me as a sanity check we should do a query and confirm that all the encounters that these dispensing constructs get migrated into are HIV dispensing encounters?

cioan:
Thanks @mogoodrich ! I have already checked the HIV dispensing encounters and they are created correctly. To test if the obs dispensing construct is created correctly I use a SQL query which I will attach it to the ticket. All the other dispensing obs are visible via the existing hiv-dispensing form.

mogoodrich:
Awesome, thanks @cioan !

mogoodrich:
I'm curious if this mysterious bug I discovered with Lab Results encounter is the same for the HIV dispensing encounters... when you edit and resave a dispensing encounter that has been migrated in, does it "disappear" from the visit?

https://pihemr.atlassian.net/browse/UHM-4854

PR #30, opened 2020-09-14
Guarantee that obs get a location_id

brandones:
From Mark's feedback on https://pihemr.atlassian.net/browse/UHM-4797

### DIFF
--- a/etl/src/main/java/org/pih/hivmigration/etl/sql/ProcedureSetup.groovy
+++ b/etl/src/main/java/org/pih/hivmigration/etl/sql/ProcedureSetup.groovy
@@ -30,7 +30,7 @@ class ProcedureSetup extends SqlMigrator {
                     value_coded, value_numeric, value_datetime, value_text, creator, date_created, voided, uuid
                 )
                 SELECT
-                    o.obs_id, p.person_id, e.encounter_id, o.obs_group_id, e.encounter_date, e.location_id, q.concept_id,
+                    o.obs_id, p.person_id, e.encounter_id, o.obs_group_id, e.encounter_date, ifnull(e.location_id, 1), q.concept_id,
                     a.concept_id, o.value_numeric, o.value_datetime, o.value_text, 1, e.date_created, 0, uuid()
                 FROM tmp_obs o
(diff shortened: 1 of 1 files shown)


etl/src/main/java/org/pih/hivmigration/etl/sql/ProcedureSetup.groovy
```diff
@@ -30,7 +30,7 @@ class ProcedureSetup extends SqlMigrator {
                     value_coded, value_numeric, value_datetime, value_text, creator, date_created, voided, uuid
                 )
                 SELECT
-                    o.obs_id, p.person_id, e.encounter_id, o.obs_group_id, e.encounter_date, e.location_id, q.concept_id,
+                    o.obs_id, p.person_id, e.encounter_id, o.obs_group_id, e.encounter_date, ifnull(e.location_id, 1), q.concept_id,
```
mogoodrich, line 33: Huh... I'd have to see the original code I was reviewing again, but the issue as I saw it was that an obs was not getting a location_id set even though it's parent encounter had a location, which isn't what this seems to address?

Doing a quick query on Haiti HIV Test I see:

mysql> select count(*) from encounter where location_id is null;
+----------+
| count(*) |
+----------+
|        0 |
+----------+
1 row in set (0.01 sec)

mysql> select count(*) from obs where location_id is null and encounter_id is not null;
+----------+
| count(*) |
+----------+
|  1441514 |
+----------+
1 row in set (2.54 sec)

mogoodrich (changes requested):

etl/src/main/java/org/pih/hivmigration/etl/sql/ProcedureSetup.groovy
brandones: Yeah but `null`s in the location_id column of `hivmigration_encounter` become `1` on the way into `encounter`. `obs` is getting its location_id from the `hivmigration_encounter` table, so that coalescing hasn't happened yet.

mogoodrich: Ah, got it, thanks!

mogoodrich (approved):

PR #31, opened 2020-09-15
Fix typo in program migrator, about outcome and outcome_date

brandones:
Relates to https://pihemr.atlassian.net/browse/UHM-4815


mseaton (approved):
Nice catch - LGTM

mogoodrich (approved):
Haven't been following the details of the program migration fully, but this definitely looks like a bug and the proper fix!

PR #32, opened 2020-09-17
UHM-4801, migrate next exam date

cioan:
Hi @mseaton , here is the migration script for NEXT_EXAM_DATE. Thanks!

### DIFF
--- a/etl/src/main/java/org/pih/hivmigration/etl/sql/ExamExtraMigrator.groovy
+++ b/etl/src/main/java/org/pih/hivmigration/etl/sql/ExamExtraMigrator.groovy
@@ -1,6 +1,6 @@
 package org.pih.hivmigration.etl.sql
 
-class ReturnVisitDateMigrator extends SqlMigrator{
+class ExamExtraMigrator extends SqlMigrator{
 
     @Override
     def void migrate() {
@@ -37,12 +37,11 @@ class ReturnVisitDateMigrator extends SqlMigrator{
         executeMysql("Load next visit date as observations", ''' 
                       
             SET @next_visit_date_concept_uuid = '3ce94df0-26fe-102b-80cb-0017a47871b2';            
-            
-            -- Do not overwrite existing Next Dispense Date obs that were inserted while migrating meds
+                        
             INSERT INTO tmp_obs (value_datetime, source_patient_id, source_encounter_id, concept_uuid)
             SELECT next_exam_date, source_patient_id, source_encounter_id, @next_visit_date_concept_uuid
             FROM hivmigration_next_visit_date
-            WHERE next_exam_date is not null and source_encounter_id not in (select e.source_encounter_id from hivmigration_encounters e, obs o, concept c where e.encounter_id = o.encounter_id and o.concept_id=c.concept_id and c.uuid=@next_visit_date_concept_uuid);
+            WHERE next_exam_date is not null;
             
             CALL migrate_tmp_obs();
             
--- a/etl/src/main/java/org/pih/hivmigration/etl/sql/ExamExtraMigrator.groovy
+++ b/etl/src/main/java/org/pih/hivmigration/etl/sql/ExamExtraMigrator.groovy
@@ -6,7 +6,7 @@ class ExamExtraMigrator extends SqlMigrator{
     def void migrate() {
 
         executeMysql("Create staging table for migrating next scheduled visit date", '''
-            create table hivmigration_next_visit_date (                            
+            create table hivmigration_exam_extra (                            
               source_encounter_id int,
               source_patient_id int,
               next_exam_date date            
@@ -14,7 +14,7 @@ class ExamExtraMigrator extends SqlMigrator{
         ''')
 
         loadFromOracleToMySql('''
-            insert into hivmigration_next_visit_date (
+            insert into hivmigration_exam_extra (
               source_encounter_id,
               source_patient_id,
               next_exam_date 
@@ -40,7 +40,7 @@ class ExamExtraMigrator extends SqlMigrator{
                         
             INSERT INTO tmp_obs (value_datetime, source_patient_id, source_encounter_id, concept_uuid)
             SELECT next_exam_date, source_patient_id, source_encounter_id, @next_visit_date_concept_uuid
-            FROM hivmigration_next_visit_date
+            FROM hivmigration_exam_extra
             WHERE next_exam_date is not null;
             
             CALL migrate_tmp_obs();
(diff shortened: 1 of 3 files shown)


etl/src/main/java/org/pih/hivmigration/etl/sql/Migrator.java
```diff
@@ -87,6 +87,7 @@ public void main() {
             if (shouldRevert || shouldRevertOnly) {
                 revert(new MedpickupsMigrator());
                 revert(new VitalsMigrator());
+                revert(new ReturnVisitDateMigrator());
```
mseaton, line 90: @cioan - if we create a Migrator for every Obs in the system, it's going to get unmanageable.  We should be a little more thoughtful about how we create migrators for specific areas of data.  Let's think through a bit more where "Return Visit Date" falls.  I could see options like:

* TreatmentPlanMigrator - migrates everything in the last few sections of the intake/followup forms
* ExamExtraMigrator - migrates everything out of the Exam Extra table that is needed

For now, you can still just add the single data point migration, but let's give ourselves some room to expand an existing migration to handle more data in the domain, rather than have to create hundreds of Migrator classes.

etl/src/main/java/org/pih/hivmigration/etl/sql/ReturnVisitDateMigrator.groovy
```diff
@@ -0,0 +1,65 @@
+package org.pih.hivmigration.etl.sql
+
+class ReturnVisitDateMigrator extends SqlMigrator{
```
mseaton, line 3: Per above, can we rename this to something more generic that will allow us to add more Obs to this in the future?

etl/src/main/java/org/pih/hivmigration/etl/sql/ReturnVisitDateMigrator.groovy
```diff
@@ -0,0 +1,65 @@
...
+
+        executeMysql("Load next visit date as observations", ''' 
+                      
+            SET @next_visit_date_concept_uuid = '3ce94df0-26fe-102b-80cb-0017a47871b2';            
+            
+            -- Do not overwrite existing Next Dispense Date obs that were inserted while migrating meds
+            INSERT INTO tmp_obs (value_datetime, source_patient_id, source_encounter_id, concept_uuid)
+            SELECT next_exam_date, source_patient_id, source_encounter_id, @next_visit_date_concept_uuid
+            FROM hivmigration_next_visit_date
+            WHERE next_exam_date is not null and source_encounter_id not in (select e.source_encounter_id from hivmigration_encounters e, obs o, concept c where e.encounter_id = o.encounter_id and o.concept_id=c.concept_id and c.uuid=@next_visit_date_concept_uuid);
```
mseaton, line 45: NO.  We don't want to add this additional where clause and possibly exclude data.  These are different data points and we need both.  If there is a risk of overlap, then we need to deal with this by altering the way we are modeling the data.

mseaton (changes requested):
@cioan See my comments.  @brandones I think I made similar comments to you in a PR or ticket or somewhere, but those must not have made it into the ticket details.

cioan:
Thanks @mseaton ! I have updated the PR with your requests. Thanks!

mseaton (changes requested):

etl/src/main/java/org/pih/hivmigration/etl/sql/ExamExtraMigrator.groovy
```diff
@@ -0,0 +1,64 @@
+package org.pih.hivmigration.etl.sql
+
+class ExamExtraMigrator extends SqlMigrator{
+
+    @Override
+    def void migrate() {
+
+        executeMysql("Create staging table for migrating next scheduled visit date", '''
+            create table hivmigration_next_visit_date (
```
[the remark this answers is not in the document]
cioan: Thanks @mseaton ! I just refactored this staging table too.

mseaton (approved):
Looks good. Thanks @cioan .

etl/src/main/java/org/pih/hivmigration/etl/sql/ExamExtraMigrator.groovy
```diff
@@ -0,0 +1,64 @@
...
+        executeMysql("Load next visit date as observations", ''' 
+                      
+            SET @next_visit_date_concept_uuid = '3ce94df0-26fe-102b-80cb-0017a47871b2';            
+                        
+            INSERT INTO tmp_obs (value_datetime, source_patient_id, source_encounter_id, concept_uuid)
+            SELECT next_exam_date, source_patient_id, source_encounter_id, @next_visit_date_concept_uuid
+            FROM hivmigration_exam_extra
+            WHERE next_exam_date is not null;
+            
+            CALL migrate_tmp_obs();
```
mogoodrich, line 46: I must have missed this one, but I assume this is a procedure that has been created to migrate tmp_obs into the obs table?

etl/src/main/java/org/pih/hivmigration/etl/sql/ExamExtraMigrator.groovy
```diff
@@ -0,0 +1,64 @@
+package org.pih.hivmigration.etl.sql
+
+class ExamExtraMigrator extends SqlMigrator{
```
mogoodrich, line 3: Ha!  Glad I read the old comments before adding my review... I was going to suggest that this would be better named "ReturnVisitDateMigrator" but then saw that you had originally named it that and @mseaton had suggested changing it to this... :) so I'm fine keeping it like this, but I'll add my reasoning here just for reference...   I do agree that we don't want to create a migrator for every data point, but until we know how we want to group things, I thought it would make sense to say exactly what this migrator does rather than using an extrapolation (ExamExtra) that we might not use going forward.  Anyway, fine as is, but if there's another data point going forward that naturally fits with ReturnVisitDate (but isn't an "Exam Extra" data point) let's not hesitate adding it to this class and renaming this to something more appropriate.  Hope that makes sense... :)

mogoodrich (approved):
One long-winded comment on my end, but LGTM and good to merge, thanks @cioan !

etl/src/main/java/org/pih/hivmigration/etl/sql/ExamExtraMigrator.groovy
mseaton: Agree entirely @mogoodrich .

etl/src/main/java/org/pih/hivmigration/etl/sql/ExamExtraMigrator.groovy
cioan: @mogoodrich , this is a generic temp table and procedure that @brandones and @mseaton worked on while migrating the lab results:
https://github.com/PIH/hivmigration/commit/331175e5584015a640dc095b2fc90563aee7b9fd#diff-82cc5f8e97422fbfca840acb367e1bcfR3

PR #33, opened 2020-09-17
UHM-4865 Log data irregularities during HIV data migration

brandones:
https://pihemr.atlassian.net/browse/UHM-4865

### DIFF
--- a/etl/src/main/java/org/pih/hivmigration/etl/sql/InfantMigrator.groovy
+++ b/etl/src/main/java/org/pih/hivmigration/etl/sql/InfantMigrator.groovy
@@ -77,6 +77,13 @@ class InfantMigrator extends SqlMigrator {
         ''')
 
         // TODO: see https://pihemr.atlassian.net/browse/UHM-4817
+        executeMysql("Note NULL names", '''
+            INSERT INTO hivmigration_data_warnings (patient_id, field_name, field_value, note)
+            SELECT person_id, IF(given_name IS NULL, 'given name', 'family name'), NULL, 'Patient missing name'
+            FROM person_name
+            WHERE (given_name IS NULL OR family_name IS NULL)
+                AND person_id in (select person_id from hivmigration_infants);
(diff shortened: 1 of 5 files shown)


etl/src/main/java/org/pih/hivmigration/etl/sql/InfantMigrator.groovy
```diff
@@ -77,6 +77,13 @@ class InfantMigrator extends SqlMigrator {
         ''')
 
         // TODO: see https://pihemr.atlassian.net/browse/UHM-4817
+        executeMysql("Note NULL names", '''
+            INSERT INTO hivmigration_data_warnings (patient_id, field_name, field_value, note)
+            SELECT person_id, IF(given_name IS NULL, 'given name', 'family name'), NULL, 'Patient missing name'
```
mseaton, line 82: Is it possible that both are null?

mseaton: Maybe rename as "Infant missing name"

etl/src/main/java/org/pih/hivmigration/etl/sql/PatientMigrator.groovy
```diff
@@ -167,6 +167,13 @@ class PatientMigrator extends SqlMigrator {
         ''')
 
         // TODO: see https://pihemr.atlassian.net/browse/UHM-4817
+        executeMysql("Note NULL names", '''
+            INSERT INTO hivmigration_data_warnings (patient_id, field_name, field_value, note)
+            SELECT person_id, IF(given_name IS NULL, 'given name', 'family name'), NULL, 'Patient missing name'
```
mseaton, line 172: Same comment here - do we want to say "given and family name" if both are null?

mseaton (approved):
Looks good to me - just some minor comments.  Great idea!

PR #35, opened 2020-09-18
Clean up program migrator

brandones:
https://pihemr.atlassian.net/browse/UHM-4815

### DIFF
--- a/etl/src/main/java/org/pih/hivmigration/etl/sql/ProgramMigrator.groovy
+++ b/etl/src/main/java/org/pih/hivmigration/etl/sql/ProgramMigrator.groovy
@@ -138,7 +138,15 @@ class ProgramMigrator extends SqlMigrator {
                 SET outcome_date = null WHERE outcome is null;
         ''')
 
-        // TODO: log cases where art_start_date > outcome_date
+        executeMysql("Note where art_start_date > outcome_date", '''
+            INSERT INTO hivmigration_data_warnings (patient_id, field_name, field_value, note)
+            SELECT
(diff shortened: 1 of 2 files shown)


etl/src/main/java/org/pih/hivmigration/etl/sql/ProgramMigrator.groovy
```diff
@@ -138,8 +138,15 @@ class ProgramMigrator extends SqlMigrator {
                 SET outcome_date = null WHERE outcome is null;
         ''')
 
-        // TODO: If the above logic does not result in an outcome date for all outcomes, research individual cases and adjust
-        // TODO: investigate cases where art_start_date > outcome_date
+        executeMysql("Note where art_start_date > outcome_date", '''
+            INSERT INTO hivmigration_data_warnings (patient_id, field_name, field_value, note)
+            SELECT
+                p.person_id, 'art_start_date', hpr.art_start_date, CONCAT('Patient has ART start date(s) after patient outcome date of ', hpr.outcome_date, '. However, ART start dates were not migrated from the HIV EMR at all anyway.')
```
mseaton, line 144: We don't actually know whether we are going to migrate art start dates yet, this is a derived field that comes from the regimen data we have pending.  All you can say here is "Patient has XYZ after their outcome date", where XYZ could be a number of different things in the system that should not happen after an outcome - an intake or followup encounter, or anything in hiv_regimes_real with start_date > outcome date (maybe others).

mseaton (approved):

etl/src/main/java/org/pih/hivmigration/etl/sql/ProgramMigrator.groovy
mogoodrich: Not following everything here, but having a data warning about this seems good to me!

mogoodrich (approved):

PR #37, opened 2020-09-25
UHM-4832 Migration of treatment obs from HIV EMR to OpenMRS

brandones:
https://pihemr.atlassian.net/browse/UHM-4832

This only accounts for "20. Treatment / Phrophylactic measures" through "22. Treatment anti-TB"

index

Documented https://pihemr.atlassian.net/wiki/spaces/ZL/pages/1052770307/ZL+HIV+Meds+Migration+Plan

ObsLoading takes ~10 minutes

### DIFF
--- a/etl/src/main/java/org/pih/hivmigration/etl/sql/TreatmentObsMigrator.groovy
+++ b/etl/src/main/java/org/pih/hivmigration/etl/sql/TreatmentObsMigrator.groovy
@@ -1,0 +1,329 @@
+package org.pih.hivmigration.etl.sql
+
+class TreatmentObsMigrator extends SqlMigrator {
+
+    @Override
+    def void migrate() {
+        // the hivmigration_observations table gets created by ObsLoadingMigrator
+
+        executeMysql("Create staging table for ordered_other", '''
+            CREATE TABLE hivmigration_ordered_other (
+              ordered_other_id int primary key auto_increment,
+              source_encounter_id int,
+              ordered varchar(100),
+              comments varchar(4000),
+              KEY `source_encounter_idx` (`source_encounter_id`)
+            );
+        ''')
+
+        loadFromOracleToMySql('''
+            INSERT INTO hivmigration_ordered_other (source_encounter_id, ordered, comments)
+            VALUES (?, ?, ?)
+        ''', '''
+            SELECT encounter_id, ordered, comments
+            FROM hiv_ordered_other
+        ''')
+
+        executeMysql("CALL create_tmp_obs_table();")
+        setAutoIncrement("tmp_obs", "(SELECT max(obs_id) + 1 FROM obs)")
+
+
+        // doesn't appear in the OpenMRS UI but is meaningful data
+        executeMysql("Add Prophylaxis Not Indicated to tmp obs table", '''
+            INSERT INTO tmp_obs (source_encounter_id, concept_uuid, value_coded_uuid)
+            SELECT
+                source_encounter_id,
+                concept_uuid_from_mapping('PIH', '2740'),
+                concept_uuid_from_mapping('CIEL', '1066')
+            FROM hivmigration_observations
+            WHERE observation = 'prophylaxis_not_indicated' AND value = 't';
+        ''')
+
+        executeMysql("Add Prophylaxis: Yes", '''
+            INSERT INTO tmp_obs (source_encounter_id, concept_uuid, value_coded_uuid)
+            SELECT
+                source_encounter_id,
+                concept_uuid_from_mapping('PIH', '2740'),
+                concept_uuid_from_mapping('CIEL', '1065')
+            FROM hivmigration_ordered_other
+            WHERE ordered IN (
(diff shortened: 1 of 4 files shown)


cioan (approved):
Thanks @brandones ! LGTM!

mogoodrich (approved):
Generally LGTM, though I didn't review the actual migration queries in any detail.  It seems like the best way for reviewing all those would be the task of creating example data of all permutations in the source HIV TEST system, doing them migration, and then reviewing in the destination OpenMRS instance. A very tedious task, but probably necessary?

Also would be good to get @mseaton 's high level review before merging as he is closest to this.

etl/src/main/java/org/pih/hivmigration/etl/sql/TreatmentObsMigrator.groovy
```diff
@@ -0,0 +1,327 @@
...
+        executeMysql("Load intermediate prophylaxis table", '''
+            DROP PROCEDURE IF EXISTS load_prophylaxis;
+            DELIMITER $$ ;
+            CREATE PROCEDURE load_prophylaxis(_txName varchar(80))
+            BEGIN
+                INSERT INTO hivmigration_prophylaxis
+                    (name, start, cont, stopped, other_value, source_encounter_id, encounter_date)
+                SELECT
+                    _txName,
+                    IF(MAX(base.ordered = _txName), TRUE, FALSE) AS start,
```
mseaton, line 83: Can you explain what these "MAX" functions are doing?  Doesn't seem right.

etl/src/main/java/org/pih/hivmigration/etl/sql/TreatmentObsMigrator.groovy
```diff
@@ -0,0 +1,327 @@
...
+                    other.comments AS other_value,
+                    base.source_encounter_id,
+                    he.encounter_date
+                FROM hivmigration_ordered_other base
+                JOIN (
+                    SELECT source_encounter_id, comments
+                    FROM hivmigration_ordered_other
+                    WHERE ordered = 'other_prophylaxis') other ON base.source_encounter_id = other.source_encounter_id
+                JOIN hivmigration_encounters he ON base.source_encounter_id = he.source_encounter_id
+                WHERE base.ordered IN (_txName, CONCAT(_txName, '_continue'))
```
mseaton, line 95: What is this where clause for that checks on _continue?

etl/src/main/java/org/pih/hivmigration/etl/sql/TreatmentObsMigrator.groovy
```diff
@@ -0,0 +1,327 @@
...
+            FROM hivmigration_arv_regimen;
+
+            INSERT INTO tmp_obs (source_encounter_id, concept_uuid, value_coded_uuid, obs_group_id)
+            SELECT
+                source_encounter_id,
+                concept_uuid_from_mapping('CIEL', '1282'),
+                CASE comments
+                    WHEN 'azt_3tc_efv' THEN concept_uuid_from_mapping('CIEL', '160124')
+                    WHEN 'azt_3tc_nvp' THEN concept_uuid_from_mapping('CIEL', '1652')
+                    ELSE concept_uuid_from_mapping('CIEL', '5424')  -- other
```
mseaton, line 240: This doesn't seem right.  We have lots more coded values.  We should map them, even if they are not on the current form.

mseaton (commented):
Really hard to comprehensively review.  We need some better actual test cases, running against source and target, that can allow us to validate a bit more.

mogoodrich:
Good point @mseaton ... would be interesting to see if we could create some test data (or "harvest" some good test cases from teh HIV EMR) and create some actual test cases around this?

etl/src/main/java/org/pih/hivmigration/etl/sql/TreatmentObsMigrator.groovy
brandones: Yeah, this is a pattern for checking whether any member of the group matches the condition. I guess `TRUE > FALSE`.

etl/src/main/java/org/pih/hivmigration/etl/sql/TreatmentObsMigrator.groovy
brandones: It says "create a row in the new prophylaxis form if either 'initée' or 'continuée' are checked in the old form"

etl/src/main/java/org/pih/hivmigration/etl/sql/TreatmentObsMigrator.groovy
brandones: Ok, I added the other codes that exist in non-trivial quantities

Screenshot from 2020-09-29 13-48-27

brandones:
Holding off until I get iterative obs loading working...

brandones:
Waiting to merge until iterative obs migration is in: https://github.com/PIH/hivmigration/pull/40

PR #38, opened 2020-09-25
Make sure visits get created for vitals

brandones:
https://pihemr.atlassian.net/browse/UHM-4800


mogoodrich (approved):
Are there any downstream migrators that rely on Visits?  If not, might make sense just to move it all the way to the end so we don't mistakenly forget to include other encounters within Visits by playing them after the Visits migration.

PR #40, opened 2020-09-30
UHM-4894 Obs migrator should iterate so that it doesn't stall out

brandones:
https://pihemr.atlassian.net/browse/UHM-4894

### DIFF
--- a/etl/src/main/java/org/pih/hivmigration/etl/sql/ObsMigrator.groovy
+++ b/etl/src/main/java/org/pih/hivmigration/etl/sql/ObsMigrator.groovy
@@ -1,0 +1,42 @@
+package org.pih.hivmigration.etl.sql
+
+abstract class ObsMigrator extends SqlMigrator {
+
+    void create_tmp_obs_table() {
+        executeMysql("Create tmp_obs table", '''
+            DROP TABLE IF EXISTS tmp_obs;
+            CREATE TABLE tmp_obs (
+                obs_id INT PRIMARY KEY AUTO_INCREMENT,
+                obs_group_id INT,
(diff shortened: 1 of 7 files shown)


etl/src/main/java/org/pih/hivmigration/etl/sql/ObsMigrator.groovy
```diff
@@ -0,0 +1,62 @@
...
+                value_coded_uuid CHAR(38),
+                value_drug_uuid CHAR(38),
+                value_datetime DATETIME,
+                value_numeric DOUBLE,
+                value_text TEXT
+            );
+        ''')
+        setAutoIncrement('tmp_obs', '(select max(obs_id)+1 from obs)')
+    }
+
```
mseaton, line 29: This is a great idea to have an ObsMigrator superclass.  It would probably make sense to actually implement the "migrate" method here, that runs create_tmp_obs_table() first, then calls a new abstract "load_tmp_obs_table()", and then runs migrate_tmp_obs().  This way, subclasses just need to implement the new "load_tmp_obs_table" method and don't even need to add those 2 method calls.  Thoughts?

mseaton (approved):
Looks good.  One comment for consideration.

etl/src/main/java/org/pih/hivmigration/etl/sql/ObsMigrator.groovy
brandones: It can be helpful to make those calls multiple times for complex migrations, in order to more easily isolate problems. I think the burden of making those calls is worth it for both the flexibility and clarity/explicitness.

mogoodrich (commented):
LGTM!

PR #41, opened 2020-10-01
UHM-4801: Migrate "Migrate Exam Extra Info" to SQL Migrator

mogoodrich:
There was one "Next Appt Date" in the Exam Extra table that linked to an accompagnateur which could potentially clash with Next Dispensed Date on the dispensing form.

This encounter was from 2007, and it didn't seem worth spending two much time worrying about it, so, for now I changed the script to *only* migrate next exam dates associated with "intake" and "followup" encounters.  All other next exam dates (currently just the 1 mentioned above) get logged in the data warnings table.


cioan (approved):
THanks @mogoodrich !

mseaton (approved):

PR #42, opened 2020-10-02
Make function concept_uuid_from_mapping deterministic. 

brandones:
Fixes https://bamboo.pih-emr.org/browse/HIVM-EMJ-91/log


brandones:
Trivial; merging

PR #45, opened 2020-10-08
UHM-4909: Updates to HIVMIGRATION_DATA_WARNINGS table

mogoodrich:
Thoughts?

I guess it could be argued that the "priority" column is premature, since I used the same priority (WARN) for all the existing warnings... but I do think that as much as we'd like *all* the issues in this table to be addressed, it's going to grow quickly and we are going to want to highlight the ones that *must* be fixed.


brandones:
Yeah the priority thing is a good idea, but it's simple enough that I think we should leave it until we have a proper use case for it.

cioan (approved):
Good idea @mogoodrich !

mseaton (approved):
Was also thinking having encounter_id would be a good idea.  LGTM!

mogoodrich:
I removed the priority stuff for now, just adding encoutner_id

PR #50, opened 2020-10-13
Improvement to treatment obs migration

brandones:
This includes some work that I thought I'd PR'ed and committed, but apparently haven't. Hence commits from Oct 2. Sorry about that. This is a very miscellaneous set of additional pieces of the Treatment Obs migration.


brandones:
Oh, I wasn't wrong, the first three changes actually had gone in. Opening a new PR.

PR #51, opened 2020-10-13
Map old TB treatments

brandones:
This is related to https://pihemr.atlassian.net/browse/UHM-4905 . It may or may not resolve it, since it hasn't been determined what's supposed to happen. This seems like a reasonable thing to do for now though.


mogoodrich (commented):
Took a look, but not much for me to weigh in on here...

brandones:
Merging because it's slightly more correct than before. It will need to be amended again.

PR #53, opened 2020-10-13
UHM-4830, do not create coded obs with null values

cioan:
@mogoodrich , this is to fix the problem with creating "Dispensed in community" coded obs with null values. Thanks!


brandones (approved):

mogoodrich (approved):
LGTM, thanks @cioan !

PR #54, opened 2020-10-14
UHM-4834: Data Quality: Review/clean up multiple encounters at different locations at the same day

mogoodrich:
https://pihemr.atlassian.net/browse/UHM-4834


etl/src/main/java/org/pih/hivmigration/etl/sql/Setup.groovy
```diff
@@ -21,6 +21,7 @@ class Setup extends SqlMigrator {
                 id INT PRIMARY KEY AUTO_INCREMENT,
                 patient_id INT,
                 encounter_id INT,
+                encounter_date DATE,
```
mogoodrich, line 24: Encounter date was very helpful in my testing when I went to look up these issues in the existing HIV EMR.

PR #55, opened 2020-10-14
remove old Pentaho code

mogoodrich:


etl/pom.xml
```diff
@@ -61,38 +61,6 @@
...
-            </exclusions>
-            <scope>compile</scope>
-        </dependency>
-        <dependency>
-            <groupId>pentaho-kettle</groupId>
-            <artifactId>kettle-engine</artifactId>
-            <version>7.1-SNAPSHOT</version>
-            <scope>compile</scope>
-        </dependency>
-        <dependency>
```
mogoodrich, line 83: From the commits, I *think* Rhino and javax.mail were only used by the Pentaho code?

mogoodrich (commented):
I deleted the Pentaho Migrator class and then everything in the "jobs" subdirectory... we could keep the "jobs" subdirectory around (or move it to an "old" folder of some type) if we prefer.

PR #56, opened 2020-10-14
UHM-3237 A few encounters in HIV_ENCOUNTERS don't have encounter datetimes

brandones:
https://pihemr.atlassian.net/browse/UHM-3237

### DIFF
--- a/etl/src/main/java/org/pih/hivmigration/etl/sql/EncounterMigrator.groovy
+++ b/etl/src/main/java/org/pih/hivmigration/etl/sql/EncounterMigrator.groovy
@@ -118,9 +118,22 @@ class EncounterMigrator extends SqlMigrator {
                 hivmigration_patients p on e.source_patient_id = p.source_patient_id 
             left join
               hivmigration_users hu on e.source_creator_id = hu.source_user_id
-            where e.encounter_type_id is not null  -- TODO: still need to migrate 'note' and 'regime' https://pihemr.atlassian.net/browse/UHM-3244
-              and e.encounter_date is not null   -- The encounters without encounter_dates have no data
+            where e.encounter_type_id is not null  # TODO: still need to migrate 'note' and 'regime' https://pihemr.atlassian.net/browse/UHM-3244
+              and e.encounter_date is not null   # TODO: figure out what to do with these https://pihemr.atlassian.net/browse/UHM-3237
             ;
+        ''')
+
+        executeMysql("Log warnings for encounters with null encounter dates", '''
+            INSERT INTO hivmigration_data_warnings (patient_id, encounter_id, field_name, note)
+            SELECT patient_id,
+                   encounter_id,
+                   'encounter_date',
+                   CONCAT('Encounter date is null. Encounter not migrated to encounter table. ',
+                       'Source encounter_id ', source_encounter_id,
+                       '. Source patient_id ', source_patient_id,
+                       '. Source encounter_type ', source_encounter_type, '.')
+            FROM hivmigration_encounters
+            WHERE encounter_date IS NULL;
         ''')
     }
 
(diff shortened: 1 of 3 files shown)


etl/src/main/java/org/pih/hivmigration/etl/sql/EncounterMigrator.groovy
```diff
@@ -122,6 +122,19 @@ class EncounterMigrator extends SqlMigrator {
               and e.encounter_date is not null   # TODO: figure out what to do with these https://pihemr.atlassian.net/browse/UHM-3237
             ;
         ''')
+
+        executeMysql("Log warnings for encounters with null encounter dates", '''
+            INSERT INTO hivmigration_data_warnings (patient_id, encounter_id, field_name, note)
+            SELECT patient_id,
```
cioan, line 128: I am not sure why, but on my local instance the hivmigration_encounters.patient_id is null for all rows. That's why I joined this table with the hivmigration_patients to get the patient_id.

cioan (approved):
Looks good, @brandones ! I just issued a PR on the same file too. I will change the message in my recent PR to "Log warnings about ...". THanks!

etl/src/main/java/org/pih/hivmigration/etl/sql/EncounterMigrator.groovy
```diff
@@ -122,6 +122,19 @@ class EncounterMigrator extends SqlMigrator {
...
+                   encounter_id,
+                   'encounter_date',
+                   CONCAT('Encounter date is null. Encounter not migrated to encounter table. ',
```
mogoodrich, line 131: +1 for making it clear that the encounter has not been migrated.  This would probably have been my first use case for a "priority" flag, making this an "ERROR", not a warning... no need to add at this point, but might be worth considering as we go forward if we want to explicitly flag issues like these that lead to data not being migrated as higher priority than those that just point out bad data that we still are able to migrate.

mogoodrich (approved):
LGTM, thanks!

mogoodrich (commented):
Actually, let me start a thread about source_patient_id vs patient_id...

etl/src/main/java/org/pih/hivmigration/etl/sql/EncounterMigrator.groovy
brandones: Actually, none of these encounters have data associated with them, from any of
- hiv_observations
- hiv_ordered_other
- hiv_exam_extra
- hiv_exam_vital_signs
- hiv_dispensing
- hiv_dispensing_meds
- hiv_lab_results

They're probably just crap. We may not even want to warn about them. There is no data.

mogoodrich: Makes sense @brandones ... @mseaton  thoughts on if there is any other place these encounters might be used?

Agree that if they are totally bogus no need to warn about then...

cioan (approved):

etl/src/main/java/org/pih/hivmigration/etl/sql/EncounterMigrator.groovy
mogoodrich: Just pinging @mseaton again to make sure there isn't anything we might be missing where we'd need to warn about this.

Also @brandones  is the date entered for all these encounters way in the past?  (just thinking of the admittedly unlikely possibility that between now and the time of migration an encounter will be added to the system without an encounter datetime but that *does* have some important data we don't want to lose.. so would be good to see if this is an "old" error or something that is stll occurring)

mseaton: Well those aren't the only tables that foreign key to encounter.  And some encounters are valid on their own.  But I haven't done any research into this so really have no idea if these encounters might be valid or not.  Not much downside to logging them as issues.

PR #57, opened 2020-10-14
UHM-3238, log warnings for encounters with invalid dates

cioan:
@mogoodrich , here are the changes for logging encounter with invalid dates. Thanks!


mogoodrich (commented):
Actually, looking through this and @brandones 's PR I have a question about "patient_id" in this table... I'll follow up on Teams...

mogoodrich (approved):
LGTM, will create a new ticket about adding other identifier types, sorry about the waffling!

PR #58, opened 2020-10-16
UHM-4907, migrate transferred_out locations

cioan:
This is for migrating transferred_out locations. Thanks!


mseaton (changes requested):
See my comments on the ticket.  If there is historical data to capture the coded (yes/no) data on "Is this patient being transferred out", then we need to migrate this in addition to the details of where.

cioan:
Thanks @mseaton ! That was easy enough. I just updated the PR to migrate the transfer_p obs.

cioan:
@mseaton , I will merge this, but still open for post commit comments. Thanks!

PR #60, opened 2020-10-21
UHM-4907, do not migrate transfer_p obs yet

cioan:
@mogoodrich , I have removed the transfer_p obs from this migration task. Thanks for reviewing this update.


mogoodrich (approved):
LGTM, thanks @cioan

PR #62, opened 2020-10-21
ARV Tx names should be PIH:1282, grouped under PIH:6116

brandones:
https://pihemr.atlassian.net/browse/UHM-4832

I didn't test this locally, so it's on me if it breaks everything.


brandones:
@dmdesimone feel free to merge as before if that looks right.

dmdesimone:
Look good!

PR #64, opened 2020-10-22
UHM-4950 Migrate CD4 from HIV_EXAM_LAB_RESULTS

brandones:
https://pihemr.atlassian.net/browse/UHM-4950

This begins the ExamLabResults migrator.

### DIFF
--- a/etl/src/main/java/org/pih/hivmigration/etl/sql/ExamLabResultsMigrator.groovy
+++ b/etl/src/main/java/org/pih/hivmigration/etl/sql/ExamLabResultsMigrator.groovy
@@ -1,0 +1,80 @@
+package org.pih.hivmigration.etl.sql
+
+class ExamLabResultsMigrator extends ObsMigrator {
+    @Override
+    def void migrate() {
+
+        executeMysql("Create hivmigration_exam_lab_results temp table", '''
+            CREATE TABLE hivmigration_exam_lab_results (
+                source_encounter_id INT,
+                lab_test varchar(80),
+                result varchar(255),
+                test_date date
+            );
+        ''')
+
+        loadFromOracleToMySql('''
+            INSERT INTO hivmigration_exam_lab_results (
+                source_encounter_id, lab_test, result, test_date
+            ) VALUES (?, ?, ?, ?)
+        ''',
+        '''
+            SELECT encounter_id, lab_test, result, test_date FROM HIV_EXAM_LAB_RESULTS
+        ''')
+
+        create_tmp_obs_table()
+
+        executeMysql("Migrate CD4", '''
+            INSERT INTO tmp_obs (source_encounter_id, concept_uuid, value_numeric)
+            SELECT
+                he.source_encounter_id,
+                concept_uuid_from_mapping('CIEL', '5497'),
+                CASE
+                    WHEN is_number(result) THEN TRIM(result)
+                    WHEN result REGEXP '^\\d*\\s*mm$' THEN extract_number(result)
+                    END
+            FROM hivmigration_exam_lab_results helr
+            JOIN hivmigration_encounters he on helr.source_encounter_id = he.source_encounter_id
+            JOIN encounter e on he.encounter_id = e.encounter_id
+            WHERE
+                lab_test = 'cd4'    
+                AND is_number(result) OR result REGEXP '^\\d*\\s*mm$';
+        ''')
+
+        executeMysql("Log warning about invalid CD4 values", '''
+            INSERT INTO hivmigration_data_warnings (patient_id, encounter_id, field_name, field_value, note)
+            SELECT
+                hp.person_id,
+                he.encounter_id,
+                'HIV_EXAM_LAB_RESULTS cd4',
+                helr.result,
+                'Invalid value. Not migrated.'
+            FROM hivmigration_exam_lab_results helr
+            JOIN hivmigration_encounters he on helr.source_encounter_id = he.source_encounter_id
+            JOIN hivmigration_patients hp on he.source_patient_id = hp.source_patient_id
+            WHERE lab_test = 'cd4' AND NOT is_number(result) AND NOT result REGEXP '^\\d*\\s*mm$';
+        ''')
+
+        executeMysql("Log warning about missing encounters", '''
+            INSERT INTO hivmigration_data_warnings (patient_id, encounter_id, field_name, field_value, note)
+            SELECT
+                hp.person_id,
+                he.encounter_id,
+                'HIV_EXAM_LAB_RESULTS cd4',
+                helr.result,
+                CONCAT('Belongs to an encounter that was not migrated, source ID ', he.source_encounter_id, '. Not migrated')
+            FROM hivmigration_exam_lab_results helr
+                     JOIN hivmigration_encounters he on helr.source_encounter_id = he.source_encounter_id
+                     JOIN hivmigration_patients hp on he.source_patient_id = hp.source_patient_id
+                     LEFT JOIN encounter e on he.encounter_id = e.encounter_id 
+            WHERE lab_test = 'cd4' AND e.encounter_id IS NULL;
+        ''')
+
+        migrate_tmp_obs()
+    }
+
+    @Override
+    def void revert() {
+        executeMysql("DROP TABLE IF EXISTS hivmigration_exam_lab_results;")
+    }
+}
--- a/etl/src/main/java/org/pih/hivmigration/etl/sql/ExamLabResultsMigrator.groovy
+++ b/etl/src/main/java/org/pih/hivmigration/etl/sql/ExamLabResultsMigrator.groovy
@@ -19,7 +19,8 @@ class ExamLabResultsMigrator extends ObsMigrator {
             ) VALUES (?, ?, ?, ?)
         ''',
         '''
-            SELECT encounter_id, lab_test, result, test_date FROM HIV_EXAM_LAB_RESULTS
+            SELECT encounter_id, lab_test, result, test_date 
+            FROM HIV_EXAM_LAB_RESULTS
         ''')
 
         create_tmp_obs_table()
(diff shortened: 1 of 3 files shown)


cioan:
Thanks @brandones . Why did you create a separate staging table? You had already created hivmigration_lab_results:
https://github.com/PIH/hivmigration/blob/master/etl/src/main/java/org/pih/hivmigration/etl/sql/LabResultMigrator.groovy#L8

and you used it to migrate CD4 results:
https://github.com/PIH/hivmigration/blob/master/etl/src/main/java/org/pih/hivmigration/etl/sql/LabResultMigrator.groovy#L113

Why this CD4 test should go in a separate staging table and migrated using a different process?

brandones:
Oh yeah, good point, this could indeed be consolidated.

brandones:
Oh wait, that intermediate table has a totally different structure. While the hiv_exam_lab_results table just has "lab_test" and "result", the hiv_lab_results table has different columns for each result type—numeric, text, boolean. I think it will be easier to handle these results separately. Things get weird with the data normalization, i.e. doing different things based on the units etc.

etl/src/main/java/org/pih/hivmigration/etl/sql/ExamLabResultsMigrator.groovy
```diff
@@ -0,0 +1,80 @@
...
+        create_tmp_obs_table()
+
+        executeMysql("Migrate CD4", '''
+            INSERT INTO tmp_obs (source_encounter_id, concept_uuid, value_numeric)
+            SELECT
+                he.source_encounter_id,
+                concept_uuid_from_mapping('CIEL', '5497'),
+                CASE
+                    WHEN is_number(result) THEN TRIM(result)
+                    WHEN result REGEXP '^\\d*\\s*mm$' THEN extract_number(result)
```
mogoodrich, line 34: I see "is_number" defined below, but where is extract_number defined?

etl/src/main/java/org/pih/hivmigration/etl/sql/ExamLabResultsMigrator.groovy
```diff
@@ -0,0 +1,80 @@
...
+                    END
+            FROM hivmigration_exam_lab_results helr
+            JOIN hivmigration_encounters he on helr.source_encounter_id = he.source_encounter_id
+            JOIN encounter e on he.encounter_id = e.encounter_id
```
mogoodrich, line 38: I'm probably just missing it, but why are we joining on "encounter" here?  I don't see where we are using it?

etl/src/main/java/org/pih/hivmigration/etl/sql/ExamLabResultsMigrator.groovy
```diff
@@ -0,0 +1,80 @@
...
+            WHERE
+                lab_test = 'cd4'    
+                AND is_number(result) OR result REGEXP '^\\d*\\s*mm$';
```
mogoodrich, line 41: Would be nice to give an example that matches that REGEXP, since regexs are so hard to read.

etl/src/main/java/org/pih/hivmigration/etl/sql/ExamLabResultsMigrator.groovy
```diff
@@ -0,0 +1,80 @@
...
+                hp.person_id,
+                he.encounter_id,
+                'HIV_EXAM_LAB_RESULTS cd4',
+                helr.result,
+                CONCAT('Belongs to an encounter that was not migrated, source ID ', he.source_encounter_id, '. Not migrated')
+            FROM hivmigration_exam_lab_results helr
+                     JOIN hivmigration_encounters he on helr.source_encounter_id = he.source_encounter_id
+                     JOIN hivmigration_patients hp on he.source_patient_id = hp.source_patient_id
+                     LEFT JOIN encounter e on he.encounter_id = e.encounter_id 
+            WHERE lab_test = 'cd4' AND e.encounter_id IS NULL;
```
mogoodrich, line 70: Per the above, if the encounter was not migrated because it was for a test patient, we should just ignore (and don't need to add a warning).  If there's some other, unknown reason, then, yeah, we should log a warning.

etl/src/main/java/org/pih/hivmigration/etl/sql/Setup.groovy
```diff
@@ -16,6 +16,12 @@ class Setup extends SqlMigrator {
             DELIMITER ;
         ''')
 
+        executeMysql('''
+            DROP FUNCTION IF EXISTS is_number;
+            CREATE FUNCTION is_number (_value text) RETURNS boolean DETERMINISTIC
+            RETURN concat('', _value * 1) = _value;
```
mogoodrich, line 22: this is an interesting one! might want to add a link to the source for this in case anyone is doubting your logic, this one seemed a bit cryptic to me, but I googled and found the link I'm guessing you did:  

https://stackoverflow.com/questions/5064977/detect-if-value-is-number-in-mysql

:)

etl/src/main/java/org/pih/hivmigration/etl/sql/ExamLabResultsMigrator.groovy
```diff
@@ -0,0 +1,80 @@
+package org.pih.hivmigration.etl.sql
+
+class ExamLabResultsMigrator extends ObsMigrator {
```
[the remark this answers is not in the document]
brandones: Yep!

etl/src/main/java/org/pih/hivmigration/etl/sql/ExamLabResultsMigrator.groovy
```diff
@@ -0,0 +1,80 @@
...
+            );
+        ''')
+
+        loadFromOracleToMySql('''
+            INSERT INTO hivmigration_exam_lab_results (
+                source_encounter_id, lab_test, result, test_date
+            ) VALUES (?, ?, ?, ?)
+        ''',
+        '''
+            SELECT encounter_id, lab_test, result, test_date FROM HIV_EXAM_LAB_RESULTS
```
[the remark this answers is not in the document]
brandones: Ohhhh yes, I'm sure we do. Good catch.

etl/src/main/java/org/pih/hivmigration/etl/sql/ExamLabResultsMigrator.groovy
brandones: Forgot to include it, thanks!

etl/src/main/java/org/pih/hivmigration/etl/sql/ExamLabResultsMigrator.groovy
brandones: This just guarantees that we only migrate lab results corresponding to an encounter that was successfully migrated (hence the warning later). I'll add a comment.

etl/src/main/java/org/pih/hivmigration/etl/sql/ExamLabResultsMigrator.groovy
brandones: Good call!

etl/src/main/java/org/pih/hivmigration/etl/sql/Setup.groovy
brandones: That's the one, yeah, the comment is a good idea

etl/src/main/java/org/pih/hivmigration/etl/sql/ExamLabResultsMigrator.groovy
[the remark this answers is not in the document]
brandones: Wait, all this is being joined on the hivmigration_encounters table below. Leaving as-is.

mogoodrich (approved):
LGTM, thanks @brandones !

etl/src/main/java/org/pih/hivmigration/etl/sql/ExamLabResultsMigrator.groovy
```diff
@@ -0,0 +1,81 @@
...
+                CASE
+                    WHEN is_number(result) THEN TRIM(result)
+                    WHEN result REGEXP '^\\d*\\s*mm$' THEN extract_number(result)
+                    END
+            FROM hivmigration_exam_lab_results helr
+            JOIN hivmigration_encounters he on helr.source_encounter_id = he.source_encounter_id
+            JOIN encounter e on he.encounter_id = e.encounter_id  -- only migrate lab results corresponding to an encounter that was successfully migrated
+            WHERE
+                lab_test = 'cd4'    
+                AND is_number(result) OR result REGEXP '^\\d*\\s*mm$';  -- e.g. '20 mm'
```
mseaton, line 42: I don't think this is right.  CD4 counts are not meaured in "mm".  I would expect this regex would be correct for, say, ppd result, but not for a CD4 result.  If the result of a CD4 is "20 mm", this should be logged as an error, not a CD4 of 20, particularly since a CD4 of 20 is clinically relevant (and very concerning).

etl/src/main/java/org/pih/hivmigration/etl/sql/ExamLabResultsMigrator.groovy
```diff
@@ -0,0 +1,81 @@
...
+            SELECT
+                hp.person_id,
+                he.encounter_id,
+                'HIV_EXAM_LAB_RESULTS cd4',
+                helr.result,
+                'Invalid value. Not migrated.'
+            FROM hivmigration_exam_lab_results helr
+            JOIN hivmigration_encounters he on helr.source_encounter_id = he.source_encounter_id
+            JOIN hivmigration_patients hp on he.source_patient_id = hp.source_patient_id
+            WHERE lab_test = 'cd4' AND NOT is_number(result) AND NOT result REGEXP '^\\d*\\s*mm$';
```
mseaton, line 56: This seems error prone to have to maintain the same query as above, but with the inverse conditional.  It would be better to, say, just look for any rows in the hivmigration_exam_lab_results of type 'cd4' that do not have corresponding rows in tmp_obs, as this represents the actual rows not migrated, and allows us to continuously improve the migration sql without having to ensure we update the corresponding data warning statement.

mseaton: Also, same comment here about "mm" not being valid for CD4

mseaton (commented):
@brandones  / @mogoodrich - sorry, I'm late on some of my reviews. I know this has been merged, but see my comments.

etl/src/main/java/org/pih/hivmigration/etl/sql/ExamLabResultsMigrator.groovy
mogoodrich: +1 to @mseaton 's idea of doing the reverse lookup in tmp_obs.

brandones:
Good catch, and great idea, @mseaton . Thanks for the review. I'll submit another PR.

etl/src/main/java/org/pih/hivmigration/etl/sql/ExamLabResultsMigrator.groovy
brandones: https://github.com/PIH/hivmigration/pull/75

PR #67, opened 2020-10-23
UHM-4953 migrate RPR test from HIV_EXAM_LAB_RESULTS. Also, fix data warnings.

brandones:
https://pihemr.atlassian.net/browse/UHM-4953

### DIFF
--- a/etl/src/main/java/org/pih/hivmigration/etl/sql/ExamLabResultsMigrator.groovy
+++ b/etl/src/main/java/org/pih/hivmigration/etl/sql/ExamLabResultsMigrator.groovy
@@ -47,7 +47,7 @@ class ExamLabResultsMigrator extends ObsMigrator {
         ''')
 
         executeMysql("Log warning about invalid CD4 values", '''
-            INSERT INTO hivmigration_data_warnings (patient_id, encounter_id, field_name, field_value, note)
+            INSERT INTO hivmigration_data_warnings (openmrs_patient_id, openmrs_encounter_id, field_name, field_value, warning_type)
             SELECT
                 hp.person_id,
                 he.encounter_id,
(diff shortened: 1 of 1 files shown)


etl/src/main/java/org/pih/hivmigration/etl/sql/ExamLabResultsMigrator.groovy
```diff
@@ -113,7 +113,7 @@ class ExamLabResultsMigrator extends ObsMigrator {
         ''')
 
         executeMysql("Log warning about invalid PPD values", '''
-            INSERT INTO hivmigration_data_warnings (patient_id, encounter_id, field_name, field_value, note)
+            INSERT INTO hivmigration_data_warnings (openmrs_patient_id, openmrs_encounter_id, field_name, field_value, warning_type)
```
mogoodrich, line 116: iirc, the note was something like generic like "Invalid valid."  I think we'd want something like "Invalid PPD value", so our aggregrate counts (which group by "warning_type") would be by test.  You could check with Debbie and/or Ricard and see what they prefer, but I assume they will want to tackle the DQ issues in the more important tests first, so a count by test would make sense.

etl/src/main/java/org/pih/hivmigration/etl/sql/ExamLabResultsMigrator.groovy
```diff
@@ -142,6 +142,60 @@ class ExamLabResultsMigrator extends ObsMigrator {
...
+                hp.person_id,
+                he.encounter_id,
+                'HIV_EXAM_LAB_RESULTS rpr',
+                helr.result,
+                'Belongs to an encounter that was not migrated. Value not migrated.'
+            FROM hivmigration_exam_lab_results helr
+            JOIN hivmigration_encounters he on helr.source_encounter_id = he.source_encounter_id
+            JOIN hivmigration_patients hp on he.source_patient_id = hp.source_patient_id
+            LEFT JOIN encounter e on he.encounter_id = e.encounter_id
+            WHERE lab_test = 'rpr' AND e.encounter_id IS NULL;
```
mogoodrich, line 195: Following on from my comment on the previous commit... we may want to have a way to not have to write the query for every different type of lab test, and just check that every encounter in the hivmigration_encounters ends up in the encounter table... especially now if we are writing a query like this for every lab test.

Will be interesting to see what this table shows... tangentially, I think these will likely be more migration errors than DQ errors... ie things we are missing in our migration and need to fix on the migration side

mogoodrich (approved):
made a couple comments, but LGTM!

etl/src/main/java/org/pih/hivmigration/etl/sql/ExamLabResultsMigrator.groovy
brandones: Yes, good call. I'll update these.

brandones: https://github.com/PIH/hivmigration/pull/69

etl/src/main/java/org/pih/hivmigration/etl/sql/ExamLabResultsMigrator.groovy
brandones: https://github.com/PIH/hivmigration/pull/70

PR #68, opened 2020-10-26
UHM-4954, migrate hematocrite lab results

cioan:


etl/src/main/java/org/pih/hivmigration/etl/sql/ExamLabResultsMigrator.groovy
```diff
@@ -75,6 +75,48 @@ class ExamLabResultsMigrator extends ObsMigrator {
...
+        ''')
+
+        executeMysql("Log warning about missing encounters", '''
+            INSERT INTO hivmigration_data_warnings (openmrs_patient_id, openmrs_encounter_id, field_name, field_value, warning_type)
+            SELECT
+                hp.person_id,    
+                he.encounter_id,
+                'HIV_EXAM_LAB_RESULTS hematocrite',
+                helr.result,
+                'Belongs to an encounter that was not migrated. Value not migrated.'
```
brandones, line 112:
```suggestion
                'Hematocrite value belongs to an encounter that was not migrated. Value not migrated.'
```

brandones (approved):

etl/src/main/java/org/pih/hivmigration/etl/sql/ExamLabResultsMigrator.groovy
cioan: Thanks @brandones !

mogoodrich (commented):
LGTM!

PR #72, opened 2020-10-26
UHM-4969, do not migrate encounters with test patients

cioan:
Do we also need to update all the obs migration scripts to exclude test patients?

### DIFF
--- a/etl/src/main/java/org/pih/hivmigration/etl/sql/EncounterMigrator.groovy
+++ b/etl/src/main/java/org/pih/hivmigration/etl/sql/EncounterMigrator.groovy
@@ -54,8 +54,8 @@ class EncounterMigrator extends SqlMigrator {
                     e.note_title,
                     e.response_to
                 from
-                    hiv_encounters e, hiv_demographics p 
-                    where e.patient_id = p.patient_id and (p.treatment_status is null or p.treatment_status != 'test')
+                    hiv_encounters e, hiv_demographics_real p 
+                    where e.patient_id = p.patient_id 
 
             '''
         )
--- a/etl/src/main/java/org/pih/hivmigration/etl/sql/EncounterMigrator.groovy
+++ b/etl/src/main/java/org/pih/hivmigration/etl/sql/EncounterMigrator.groovy
@@ -54,7 +54,8 @@ class EncounterMigrator extends SqlMigrator {
                     e.note_title,
                     e.response_to
                 from
-                    hiv_encounters e
+                    hiv_encounters e, hiv_demographics p 
+                    where e.patient_id = p.patient_id and (p.treatment_status is null or p.treatment_status != 'test')
 
             '''
         )


etl/src/main/java/org/pih/hivmigration/etl/sql/EncounterMigrator.groovy
```diff
@@ -54,7 +54,8 @@ class EncounterMigrator extends SqlMigrator {
                     e.note_title,
                     e.response_to
                 from
-                    hiv_encounters e
+                    hiv_encounters e, hiv_demographics p 
+                    where e.patient_id = p.patient_id and (p.treatment_status is null or p.treatment_status != 'test')
```
mogoodrich, line 58: The hiv_demographics_real view takes care of this for you... ie, you just join on  "hiv_demographics_real" instead of hiv_demographics, as that is a view with all test patients excluded.

brandones, line 58: So test patients with null treatment_status should be migrated? Why?

mogoodrich (changes requested):
We should just be able to join on the hiv_demographics_real table to to take care of this for us...

etl/src/main/java/org/pih/hivmigration/etl/sql/EncounterMigrator.groovy
mogoodrich: Dueling comments!  :)  See my comment below...

mogoodrich:
@cioan  to your obs question, yeah, we should excluding those as well... generally, we should be able to join again the "hiv_demographics_real" table to exclude any encounters/obs/etc associated with test patients.  There's  few examples of where we are currently doing this, but I guess we didn't 100% establish it as a pattern.  I'll create a new ticket for this.

cioan:
Thanks @mogoodrich ! I've updated the query to join on the hiv_demographics_real view. I could take care of the new ticket you opened for updating the obs.

etl/src/main/java/org/pih/hivmigration/etl/sql/EncounterMigrator.groovy
cioan: From what I remember from a long time ago, joining on a view is much slower than joining on the real table. I think that Oracle made some improvements and it sounds like a Materialized View is much faster, but I do not think that hiv_demographics_real is a materialized view?

mogoodrich (approved):
LGTM, thanks @cioan !  And thanks for picking up the obs ticket as well!

etl/src/main/java/org/pih/hivmigration/etl/sql/EncounterMigrator.groovy
mogoodrich: Interesting point @cioan ...   Looks like the job had been taking about 150-160ish minutes to run currently.

Let's see what the time is after this addition and if it's significant look at ways to tweak.  (Since there's no join now, this would be just a comparison of no join vs the view join... but we can try to do more details tests if we find this adds a lot to the build).

fyi @mseaton

mseaton: It's certainly easy enough to test the performance of one query that joins against the view, and another that doesn't.  My personal assumption is that joining against a view is faster or the same, as the query engine knows what to expect up-front (the view has been created, even if non-matierialized) and Oracle could theoretically take indexing and other steps to optimize for it.

That said, the other approach is just to make sure we join on the hivmigration_patients or hivmigration_encounters table at the time of import into associated staging tables, since these will have already been created with the appropriate patients and encounters for migration.

We should treat those as foreign keys that need to be checked, even if those foreign key relationships don't actually exist in the staging tables.

PR #74, opened 2020-10-27
UHM-4970 migrate hemoglobin from HIV_EXAM_LAB_RESULTS

brandones:
https://pihemr.atlassian.net/browse/UHM-4970

### DIFF
--- a/etl/src/main/java/org/pih/hivmigration/etl/sql/ExamLabResultsMigrator.groovy
+++ b/etl/src/main/java/org/pih/hivmigration/etl/sql/ExamLabResultsMigrator.groovy
@@ -132,6 +132,16 @@ class ExamLabResultsMigrator extends ObsMigrator {
                     END''',
                 "result IN ('negative', 'negatif', 'positive', '+++', '++', '+')"
         )
+
+        migrateLab("hemoglobin",
+                "concept_uuid_from_mapping('PIH', 'HEMOGLOBIN')",
+                "extract_number(result)",
+                "NULL",
+                '''result REGEXP '^[[:blank:]]*[0-9\\.]+[[:blank:]]*gr?.dll?$' OR is_number(result)'''
+                // e.g. '9.8', '9.8 g/dl', or '9.8gr/dl'
+        )
+
+        migrate_tmp_obs()
     }
 
     @Override
--- a/etl/src/main/java/org/pih/hivmigration/etl/sql/Setup.groovy
+++ b/etl/src/main/java/org/pih/hivmigration/etl/sql/Setup.groovy
@@ -5,7 +5,7 @@ class Setup extends SqlMigrator {
     @Override
     def void migrate() {
 
-        executeMysql('''
+        executeMysql("Create function concept_uuid_from_mapping", '''
             DROP FUNCTION IF EXISTS concept_uuid_from_mapping;
             DELIMITER //
             CREATE FUNCTION concept_uuid_from_mapping ( _source varchar(80), _code varchar(80))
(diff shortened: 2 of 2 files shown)


etl/src/main/java/org/pih/hivmigration/etl/sql/ExamLabResultsMigrator.groovy
```diff
@@ -49,10 +49,7 @@ class ExamLabResultsMigrator extends ObsMigrator {
             SELECT
                 he.source_encounter_id,
                 concept_uuid_from_mapping('CIEL', '5497'),
-                CASE
-                    WHEN is_number(result) THEN TRIM(result)
-                    WHEN result REGEXP '^\\d*\\s*mm$' THEN extract_number(result)
-                    END
+                extract_number(result)
```
mogoodrich, line 52: Is this based on the comments from @mseaton  earlier today, or is it unrelated? If it's related, shouldn't we be changing the WHERE clause here too to skip the ones that have "mm"?

etl/src/main/java/org/pih/hivmigration/etl/sql/ExamLabResultsMigrator.groovy
```diff
@@ -243,6 +240,61 @@ class ExamLabResultsMigrator extends ObsMigrator {
...
+                'Invalid hemoglobin test value. Not migrated.'
+            FROM hivmigration_exam_lab_results helr
+            JOIN hivmigration_encounters he ON helr.source_encounter_id = he.source_encounter_id
+            JOIN hivmigration_patients hp ON he.source_patient_id = hp.source_patient_id
+            WHERE lab_test = 'hemoglobin'
+                  AND result NOT regexp '^[[:blank:]]*[0-9\\.]+[[:blank:]]*gr?.dll?$'
+                  AND result NOT regexp '^(Hte?|HTE?)?[[:blank:]]?[0-9\\.]+%$'
+                  AND NOT is_number(result)
+                  AND NOT is_number(REPLACE(result, ',', '.'));
+          ''')
```
mogoodrich, line 295: I like @mseaton 's idea from the other PR of not doing an inverse query here, but just seeing what rows didn't make it into tmp_obs.

etl/src/main/java/org/pih/hivmigration/etl/sql/ExamLabResultsMigrator.groovy
```diff
@@ -243,6 +240,61 @@ class ExamLabResultsMigrator extends ObsMigrator {
...
+            INSERT INTO tmp_obs (source_encounter_id, concept_uuid, value_numeric)
+            SELECT
+                he.source_encounter_id,
+                concept_uuid_from_mapping('PIH', 'HEMATOCRIT'),
+                extract_number(result)
+            FROM hivmigration_exam_lab_results helr
+            JOIN hivmigration_encounters he on helr.source_encounter_id = he.source_encounter_id
+            JOIN encounter e on he.encounter_id = e.encounter_id
+            WHERE lab_test = 'hemoglobin'
+                AND result REGEXP '^(Hte?|HTE?)?[[:blank:]]?[0-9\\.]+%$'  -- e.g. '33%' or 'Ht 33.0%'
```
mogoodrich, line 275: Seems a reasonable assumption to me, but don't know if we need any outside validation?

cioan (approved):
LGTM. Thanks @brandones !

etl/src/main/java/org/pih/hivmigration/etl/sql/Setup.groovy
```diff
@@ -16,16 +16,23 @@ class Setup extends SqlMigrator {
...
-            -- from https://stackoverflow.com/a/5065007/1464495
+        executeMysql("Create function is_number", '''
+            -- Checks whether a string value represents a number. Allows surrounding whitespace.
+            -- Allows an arbitrary number of decimal points or commas.
+            -- Recommended to use with 'extract_number', e.g.
+            --   SELECT extract_number(result) FROM tbl WHERE is_number(result)
             DROP FUNCTION IF EXISTS is_number;
             CREATE FUNCTION is_number (_value text) RETURNS boolean DETERMINISTIC
-            RETURN concat('', _value * 1) = _value;
+                RETURN _value REGEXP '^[0-9\\.,]+$' AND _value REGEXP '[0-9]';
```
mogoodrich, line 26: My REGEXP skills are fading :) ... what does the first match?  It looks like a NOT match  on 0-9 and , and . but I figure I'm wrong about that?

I do think it's clearer to have a REGEX than the weird concat thing...

mogoodrich (commented):
Thanks @brandones ... I had a few comments/questions, but I don't think there are any blockers here...

etl/src/main/java/org/pih/hivmigration/etl/sql/Setup.groovy
brandones: `^` means beginning. `$` means end. `^ only means "not the following" when inside and at the start of [] , so [^...]` (from SO).

So this is "any character 0-9, or '.', or ','"

etl/src/main/java/org/pih/hivmigration/etl/sql/ExamLabResultsMigrator.groovy
brandones: Unfortunately since there's no unique identifier for those rows that goes into tmp_obs, it's not so easy to do that query. I've got an improvement in the works. Coming tomorrow morning.

etl/src/main/java/org/pih/hivmigration/etl/sql/ExamLabResultsMigrator.groovy
brandones: Probably would be good— @dmdesimone or @lnball ?

brandones: The question is— given a bunch of values mostly in the range 25-40, which are given as percentages, even though they were put into the "hemoglobin" box can we assume that they're actually hematocrit?

etl/src/main/java/org/pih/hivmigration/etl/sql/ExamLabResultsMigrator.groovy
brandones: https://github.com/PIH/hivmigration/pull/77

etl/src/main/java/org/pih/hivmigration/etl/sql/ExamLabResultsMigrator.groovy
brandones: Resolved in a chat in Teams—we're just going to drop those values (they'll get picked up by the warnings table).

PR #77, opened 2020-10-28
Refactor ExamLabResultsMigrator

brandones:


mogoodrich (approved):
LGTM!

We might want to move the "Log warning about missing encounters" query here as well?  It would mean a query-per-test-type instead of a single query, but would prevent us from forgetting to add a new test to the where clause of that query?

brandones:
Good call @mogoodrich ! Done. Thanks for the review.

PR #78, opened 2020-10-29
UHM-4922: Create data warnings ETL/DW table

mogoodrich:
(add flag for review)

### DIFF
--- a/etl/src/main/java/org/pih/hivmigration/etl/sql/EncounterMigrator.groovy
+++ b/etl/src/main/java/org/pih/hivmigration/etl/sql/EncounterMigrator.groovy
@@ -103,13 +103,14 @@ class EncounterMigrator extends SqlMigrator {
 
         executeMysql("Log warnings about encounters with nonsensical encounter dates",
                 '''
-                INSERT INTO hivmigration_data_warnings (openmrs_patient_id, openmrs_encounter_id, encounter_date, field_name, field_value, warning_type)       
+                INSERT INTO hivmigration_data_warnings (openmrs_patient_id, openmrs_encounter_id, encounter_date, field_name, field_value, warning_type, flag_for_review)       
                 SELECT p.person_id as patient_id, 
                        e.encounter_id as encounter_id, 
                        e.encounter_date as encounter_date,
                        'encounter_date' as field_name,
                        e.encounter_date as field_value,
-                       'Encounter with nonsensical date' as warning_type
+                       'Encounter with nonsensical date' as warning_type,
+                       TRUE as flag_for_review
                 from hivmigration_encounters e, hivmigration_patients p
                 where (e.encounter_date < '1990-01-01' or e.encounter_date > date_add(now(), INTERVAL 5 YEAR)) 
                 and e.source_patient_id = p.source_patient_id;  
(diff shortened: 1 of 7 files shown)


etl/src/main/java/org/pih/hivmigration/etl/sql/EncounterMigrator.groovy
```diff
@@ -103,13 +103,14 @@ class EncounterMigrator extends SqlMigrator {
...
-                INSERT INTO hivmigration_data_warnings (openmrs_patient_id, openmrs_encounter_id, encounter_date, field_name, field_value, warning_type)       
+                INSERT INTO hivmigration_data_warnings (openmrs_patient_id, openmrs_encounter_id, encounter_date, field_name, field_value, warning_type, flag_for_review)       
                 SELECT p.person_id as patient_id, 
                        e.encounter_id as encounter_id, 
                        e.encounter_date as encounter_date,
                        'encounter_date' as field_name,
                        e.encounter_date as field_value,
-                       'Encounter with nonsensical date' as warning_type
+                       'Encounter with nonsensical date' as warning_type,
+                       TRUE as flag_for_review
```
mogoodrich, line 113: By setting "TRUE" here we are basically saying "a data officer needs to look at these error at this point"... for starters, I'd think we'd just wholesale apply the flag to an warning_type...

etl/src/main/java/org/pih/hivmigration/etl/sql/Setup.groovy
```diff
@@ -75,7 +75,8 @@ class Setup extends SqlMigrator {
                 field_name VARCHAR(100),
                 field_value VARCHAR(1000),
                 warning_type VARCHAR(255),
-                warning_details VARCHAR(1000)
+                warning_details VARCHAR(1000),
+                flag_for_review BOOLEAN DEFAULT FALSE
```
cioan, line 79: Maybe it makes more sense to make flag_for_review DEFAULT TRUE, so you won't have to update all those insert statements above. I think that by default any row insert into this table should be reviewed. The use cases where this flag would be FALSE are less likely and then you could just specify it in the insert statement.

cioan (approved):
Just had one comment, but it is not a blocker.

mseaton (approved):

etl/src/main/java/org/pih/hivmigration/etl/sql/Setup.groovy
mogoodrich: Thanks @cioan !  My thought for this column was to distinguish data that was "ready for review" by a data clerk vs stuff that still needed review from us, so defaulting it to false was intended so that we would have to "manually" flag something for review by the team.

I wasn't intending for this column at all, but as I was ready to move the this ticket into test I wanted to ask Debbie to assign it to Ricard to review, and realized that there were many rows that it really didn't make sense to ask MEQ or data clerks to review yet... for example there were 3,199 "invalid" CD4 counts row, and I figured with that many we should see if there's any automated cleanup we could do first.  Then there were ones like "cd4 value belongs to an encounter that was not migrated. Value not migrated." which were really more warning for us than data clerks.  My intent was to use this field to distinguish those two types.  Hope that makes sense!

PR #79, opened 2020-10-29
do not migrate hiv_exam_lab_results entries with null values

mogoodrich:


mogoodrich:
Would like to "fast-track" this and get it in the build tonight, as it will make testing the varies lab result tickets easier...

cioan (approved):
Thanks @mogoodrich !

brandones (approved):

mseaton (commented):
No problem on my end with these changes for now.  As long as we are highlighting somewhere (as a TODO, ticket, or data warning) that these need to be reviewed.  If a result is null, but a test date is not, one might imagine that this represents a specimen collection data for which results are not yet available.

mogoodrich:
@mseaton  would you be comfortable just completely ignoring the ones with result=null and test_date = null (there are over 700,000 of these)...

mogoodrich:
Tweaks this so it only excludes results with result=null and test_date=null, and then updated the data warnings so that these values should be logged to the data warnings...

PR #80, opened 2020-10-29
Fix TB mappings

brandones:
https://pihemr.atlassian.net/browse/UHM-4966


mogoodrich (commented):
I haven't been following all the concept discussions, so not much for me to comment on here... I assume mainly for @lnball  to confirm the concepts look correct.

PR #84, opened 2020-11-05
ExamLabResults that have test_date should use that for the obs.obs_datetime

brandones:
Response to feedback on https://pihemr.atlassian.net/browse/UHM-4950


mogoodrich (approved):
LGTM, thanks @brandones !

PR #86, opened 2020-11-06
UHM-4982, exclude children of test patients

cioan:
This PR address both UHM-4982(exclude test infants) and UHM-4981(infants with a patient record). Thanks!

### DIFF
--- a/etl/src/main/java/org/pih/hivmigration/etl/sql/InfantMigrator.groovy
+++ b/etl/src/main/java/org/pih/hivmigration/etl/sql/InfantMigrator.groovy
@@ -40,7 +40,30 @@ class InfantMigrator extends SqlMigrator {
                     i.LAST_NAME, 
                     upper(i.GENDER), 
                     i.BIRTH_DATE  			
-                from HIV_INFANTS i
+                from HIV_INFANTS i 
+                where (i.patient_id is null) and ((i.MOTHER_PATIENT_ID is null) or (
+                    i.MOTHER_PATIENT_ID is not null 
+                    and i.MOTHER_PATIENT_ID in (select patient_id from HIV_DEMOGRAPHICS_REAL))
+                ); 
+        ''')
+
+        loadFromOracleToMySql('''
+                insert into hivmigration_data_warnings (
+                    hivemr_v1_id,
+                    field_name,
+                    field_value,
+                    warning_type,
+                    flag_for_review
+                ) values (?, ?, ?, ?, ?)
+            ''', '''
+                select 
+                    i.INFANT_ID,
+                    'patient_id',
+                    i.PATIENT_ID, 
+                    'Infant already has a patient_id record',
+                    1 			
+                from HIV_INFANTS i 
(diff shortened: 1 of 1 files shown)


etl/src/main/java/org/pih/hivmigration/etl/sql/InfantMigrator.groovy
```diff
@@ -40,7 +40,30 @@ class InfantMigrator extends SqlMigrator {
...
+                from HIV_INFANTS i 
+                where (i.patient_id is null) and ((i.MOTHER_PATIENT_ID is null) or (
+                    i.MOTHER_PATIENT_ID is not null 
+                    and i.MOTHER_PATIENT_ID in (select patient_id from HIV_DEMOGRAPHICS_REAL))
+                ); 
+        ''')
+
+        loadFromOracleToMySql('''
+                insert into hivmigration_data_warnings (
+                    hivemr_v1_id,
```
mogoodrich, line 52: I don't believe the INFANT_ID is the HIVEMR_V1_ID... (I don't think infants have HIVEMR_V1_IDs).

Instead of inserting data into the hivemr_v1_id column I would insert the source_patient_id into the source_patient_id column... then, if you look in the DataWarningMigrator, using the source_patient_id it should automatically pull in any other identifier it finds for that patient.

Then instead of field name/field value of "patient id", I'd store the "infant_id"  in the field_name/field_value fields...

mogoodrich (changes requested):
Thanks @cioan , looks good, but one suggested changed to the data warnings logging.... thanks!

etl/src/main/java/org/pih/hivmigration/etl/sql/InfantMigrator.groovy
cioan: Thanks @mogoodrich ! I have updated the field_name/field_value, but the hivmigration_data_warnings table has not source_patient_id column?
https://github.com/PIH/hivmigration/blob/master/etl/src/main/java/org/pih/hivmigration/etl/sql/Setup.groovy#L66

So, I am not sure where to insert the source_patient_id value?

mogoodrich: Ah, good point, I was totally wrong there... you'd want to insert the *OpenMRS* patient_id into the "openmrs_patient_id" column... but you don't have access to from the Oracle DB, of course.

If this were a really critical warning, we good totally rework it to be generated from the OpenMRS tables, etc, but it likely isn't a critical warning... :+1: 

I would just not insert a hivemr_v1_id at all, but rather had something to the warning_message field like "infant with infant_id 1 already has patient record with patient_id 2, skipping importing infant record"

We can tweak later if we determine it becomes important.  Thanks!

cioan: Thanks @mogoodrich ! I have updated the PR.

mogoodrich (approved):

PR #87, opened 2020-11-10
UHM-5013 Update HIV regimen treatment (Rx) for Data Migration

brandones:
https://pihemr.atlassian.net/browse/UHM-5013


mogoodrich (approved):
Generally looks good to me... I'm not 100% up to speed on all the current constructs we are using, so will likely be easier for me to review via testing once it's merged.

PR #94, opened 2020-11-18
UHM-5020, de-identify middle name

cioan:


etl/src/main/java/org/pih/hivmigration/etl/sql/DeIdentifyMigrator.groovy
```diff
@@ -158,6 +158,64 @@ class DeIdentifyMigrator extends SqlMigrator {
...
+                    WHEN p.person_id % 19 = 14 THEN 'Alba'
+                    WHEN p.person_id % 19 = 15 THEN 'Daniela'
+                    WHEN p.person_id % 19 = 16 THEN 'Martina'
+                    WHEN p.person_id % 19 = 17 THEN 'Maria'
+                    WHEN p.person_id % 19 = 18 THEN 'Lucia'
+                END)
+            where pn.middle_name is not null and pn.middle_name != 'UNKNOWN' 
+                and pn.person_id = p.person_id and p.gender = 'F' 
+                and (p.person_id in (select person_id from hivmigration_patients) 
+                    || p.person_id in (select person_id from hivmigration_infants));
```
brandones, line 189: Thought maybe this `||` was another way to write `OR` in SQL, but Google tells me that, bizarrely, it's a concatenation operator: https://stackoverflow.com/a/23919424/1464495

etl/src/main/java/org/pih/hivmigration/etl/sql/DeIdentifyMigrator.groovy
```diff
@@ -158,6 +158,64 @@ class DeIdentifyMigrator extends SqlMigrator {
                     WHEN p.source_infant_id % 19 = 18 THEN 'Lucas'
                 END)
             where given_name != 'UNKNOWN' and pn.person_id = p.person_id and (p.gender != 'F' or p.gender is null);
+            
+            
+            update person_name pn, person p
```
brandones, line 163: Can we move all this up under the person name de-identification, instead of being after the infant name de-identification?

etl/src/main/java/org/pih/hivmigration/etl/sql/DeIdentifyMigrator.groovy
```diff
@@ -158,6 +158,64 @@ class DeIdentifyMigrator extends SqlMigrator {
...
+            set pn.middle_name =
+                (CASE
+                    WHEN p.person_id % 19 = 0 THEN 'Charlotte'
```
brandones, line 166: Let's do these middle names modulo 21 so that we get minimal repeats (21 is coprime to 20 and 19). You just have to pick two more names. Or go modulo 17 and drop two names. Either way.

brandones (changes requested):
as above

cioan:
Thanks @brandones ! Makes sense. I have updated the PR with your comments. Thanks!

mogoodrich (approved):
LGTM, thanks @cioan !

cioan:
Thanks @brandones and @mogoodrich !

PR #96, opened 2020-11-20
UHM-5028 Migrate HIV_INTAKE_FORMS table to OpenMRS

brandones:
Creates generic table stager to move tables from Oracle to MySQL

### DIFF
--- a/etl/src/main/java/org/pih/hivmigration/etl/sql/TableStager.groovy
+++ b/etl/src/main/java/org/pih/hivmigration/etl/sql/TableStager.groovy
@@ -1,0 +1,60 @@
+package org.pih.hivmigration.etl.sql
+
+import org.apache.commons.dbutils.handlers.MapListHandler
+
+import java.util.stream.Collectors;
+
+class TableStager extends SqlMigrator {
+
+    String oracleTableName
+    String mysqlTableName
+
+    TableStager(String oracleTableName) {
+        this.oracleTableName = oracleTableName
+        this.mysqlTableName = "hivmigration_" + oracleTableName.replace("HIV_", "").toLowerCase()
+    }
+
+    @Override
+    void migrate() {
+        def datatypeMap = [VARCHAR2: "TEXT", NUMBER: "NUMERIC"]
+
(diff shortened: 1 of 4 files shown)


etl/src/main/java/org/pih/hivmigration/etl/sql/TableStager.groovy
```diff
@@ -0,0 +1,60 @@
...
+    String mysqlTableName
+
+    TableStager(String oracleTableName) {
+        this.oracleTableName = oracleTableName
+        this.mysqlTableName = "hivmigration_" + oracleTableName.replace("HIV_", "").toLowerCase()
+    }
+
+    @Override
+    void migrate() {
+        def datatypeMap = [VARCHAR2: "TEXT", NUMBER: "NUMERIC"]
```
mseaton, line 19: Why map varchar2 to text?  I would think we could construct an actual varchar column of the appropriate length with a bit more code.  If there is no performance or other negative impact to using text, I guess that's fine, but I would assume there is.

mseaton (approved):
Looks like a nice and helpful class.  A few additional utilities I can think of:
* The ability to exclude columns that are not relevant or used
* The ability to convert columns named *_p and have values of 't' or 'f' into type boolean 1/0
Or not.  We can use this for the cases where we just want to move an entire table over, and choose to do something custom where we need it.

mogoodrich:
I did a quick glance at this but in the interest of divide and conquer I didn't dive in too deeply because looks like Mike has already done so.  But conceptually this seems like a good model, and will allow us to systematically include things like filtering out test patients/encounters like you did.

cioan (approved):

etl/src/main/java/org/pih/hivmigration/etl/sql/TableStager.groovy
brandones: The only negative impact of using TEXT is that you can't index on it. Is that okay?

brandones: https://stackoverflow.com/a/25301046/1464495

PR #105, opened 2020-12-02
UHM-5038, migrate adverse events

cioan:


mogoodrich (commented):
Generally LGTM... though do we need to set an obs for the "Adverse event?" checkbox on the new form if the patient has one or more adverse events?

cioan:
Thanks @mogoodrich ! I have added the yes checkbox for when the adverse event is present. Thanks!

PR #107, opened 2020-12-05
UHM-5042, migrate CIVIL_STATUS

cioan:


etl/src/main/java/org/pih/hivmigration/etl/sql/RegistrationMigrator.groovy
```diff
@@ -268,15 +295,51 @@ class RegistrationMigrator extends SqlMigrator {
...
+                    p.person_id, 
+                    r.encounter_id, 
+                    r.encounter_date as obs_datetime, 
+                    1 as locationId, 
+                    @civil_status_concept as concept_id,  
+                    CASE s.civil_status 
+                        WHEN 'DIVORCED' THEN concept_from_mapping('PIH', 'DIVORCED') 
+                        WHEN 'MARRIED' THEN concept_from_mapping('PIH', 'MARRIED') 
+                        WHEN 'PARTNER' THEN concept_from_mapping('PIH', 'LIVING WITH PARTNER') 
+                        WHEN 'PLACE' THEN concept_from_mapping('PIH', 'LIVING WITH PARTNER')
```
mogoodrich, line 324: Do we need to confirm/have we confirmed that "PLACE" means "Living with Partner"?

mogoodrich (approved):
One small question, but generally LGTM...

etl/src/main/java/org/pih/hivmigration/etl/sql/RegistrationMigrator.groovy
cioan: Confirmed.

brandones (approved):

PR #108, opened 2020-12-08
UHM-5042, migrate all socio_economics data

cioan:
@mogoodrich , here is the migration of all fields from the socioeconomics table. Thanks!


mogoodrich (approved):
LGTM, thanks @cioan !

PR #111, opened 2020-12-10
UHM-4957, do not create empty NIF and NATIONAL IDs

cioan:


mogoodrich (approved):
LGTM, thanks @cioan !

PR #115, opened 2020-12-11
UHM-5046, update data type mapping

cioan:
I had to update the data mapping to be a bit more precise. Before this change, the ORACLE CHAR(8) data type was mapped to MySQL CHAR which caused the migration to fail because it could not insert 8 chars strings into 1 char field. Now, I added the DATA_LENGTH when defining the MySQL data type.


mogoodrich (commented):
LGTM, thanks @cioan !

cioan:
Thanks @mogoodrich !

PR #123, opened 2020-12-17
Deduplicate patient diagnoses in migration

brandones:
https://pihemr.atlassian.net/browse/UHM-5036

Also marks "yes" for other diagnoses.

Screenshot from 2020-12-17 15-48-42


mogoodrich (approved):
Interesting, LGTM @brandones ... do you have a good example of an encounter with duplicate data in the source?  I'd like to poke around a bit to explore how it may have happened.

brandones:
Verified with `-r -l=100000` run.

PR #125, opened 2020-12-19
Restore and fix: Create registration encounter for each infant

brandones:
This reverts commit fe6ddb5ea1ff2a0dc74fb6fe0dc2ac9a8ace719a, restoring the original commit.

The problem was that the Location migrator was running after the Infant migrator.


mogoodrich (approved):
LGTM, thanks @brandones !

PR #126, opened 2020-12-19
Fixes for UHM-5026 Fix prophylaxis and TB migration to support HIV Status section

brandones:
https://pihemr.atlassian.net/browse/UHM-5026

### DIFF
--- a/etl/src/main/java/org/pih/hivmigration/etl/sql/Setup.groovy
+++ b/etl/src/main/java/org/pih/hivmigration/etl/sql/Setup.groovy
@@ -62,6 +62,22 @@ class Setup extends SqlMigrator {
             DELIMITER ;
         ''')
 
+        executeMysql("Create function try_to_fix_date", '''
+            DROP FUNCTION IF EXISTS try_to_fix_date;
+            DELIMITER |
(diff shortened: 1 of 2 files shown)


etl/src/main/java/org/pih/hivmigration/etl/sql/Setup.groovy
```diff
@@ -62,6 +62,22 @@ class Setup extends SqlMigrator {
             DELIMITER ;
         ''')
 
+        executeMysql("Create function try_to_fix_date", '''
```
mogoodrich, line 65: Was the bug I was seeing caused by invalid date values?

mogoodrich (approved):

etl/src/main/java/org/pih/hivmigration/etl/sql/Setup.groovy
brandones: I doubt it. I expect it was caused by a faulty join. But using this function allowed simplifying the date logic substantially, eliminating those joins, and the problem.

brandones:
Verified with a successful `-r -l=200000` run.

PR #130, opened 2020-12-22
UHM-5065, migrate opportunistic infections

cioan:


etl/src/main/java/org/pih/hivmigration/etl/sql/HivExamOisMigrator.groovy
```diff
@@ -0,0 +1,248 @@
...
+            values(?,?,?,?) 
+            ''', '''
+            select 
+                x.ENCOUNTER_ID as source_encounter_id, 
+                e.patient_id as source_patient_id, 
+                lower(trim(x.oi)) as oi,                  
+                x.COMMENTS
+            from HIV_EXAM_OIS x, HIV_ENCOUNTERS e, hiv_demographics_real d  
+            where x.ENCOUNTER_ID = e.ENCOUNTER_ID and e.patient_id = d.patient_id;
+            ''')
```
brandones, line 113: You don't have to, but you can use `(new TableStager("HIV_EXAM_OIS")).migrate()` in place of all this. Or just add it to the TableStagingMigrator.

brandones: Oh, though you wouldn't then be able to use `obs_id` as the obs group ids. Unless you were to do something like

```
        Object nextId = this.selectMysql(select, new ScalarHandler());
        String sql = "ALTER TABLE " + table + " ADD COLUMN obs_id int primary key auto_increment, auto_increment=" + (nextId != null ? nextId.toString() : "1")
        executeMysql(sql)
```

cioan: Thanks @brandones ! And, then I would have to add source_patient_id. And the TYPE mapping is not quite correctly. Using TEXT for all VARCHAR makes things very, very slow. The insertion is much faster when I use smaller data types. Thanks!

etl/src/main/java/org/pih/hivmigration/etl/sql/HivExamOisMigrator.groovy
```diff
@@ -0,0 +1,248 @@
...
+            SELECT 
+                x.obs_id,
+                x.source_patient_id,
+                x.source_encounter_id,
+                concept_uuid_from_mapping('PIH', 'DIAGNOSIS') as concept_uuid,
+                concept_uuid_from_mapping(m.openmrs_concept_source, m.openmrs_concept_code) as value_coded_uuid,
+                x.comments
+            from hivmigration_hiv_exam_ois x, hivmigration_hiv_ois_mapping m  
+            where x.oi is not null and x.oi != 'none' and x.oi != 'other' and x.oi = m.oi 
+                and concept_uuid_from_mapping(m.openmrs_concept_source, m.openmrs_concept_code) is not null;
```
brandones, line 201: This seems like it could hide errors we'd want to be aware of. How about

```suggestion
                and m.openmrs_concept_code != '';  
```

etl/src/main/java/org/pih/hivmigration/etl/sql/HivExamOisMigrator.groovy
```diff
@@ -0,0 +1,248 @@
...
+            SELECT 
+                x.obs_id,
+                x.source_patient_id,
+                x.source_encounter_id,
+                concept_uuid_from_mapping('PIH', 'Diagnosis or problem, non-coded') as concept_uuid,
+                x.oi,
+                x.comments
+            from hivmigration_hiv_exam_ois x, hivmigration_hiv_ois_mapping m  
+            where x.oi is not null and x.oi != 'none' and x.oi != 'other' and x.oi = m.oi 
+                and concept_uuid_from_mapping(m.openmrs_concept_source, m.openmrs_concept_code) is null;
```
brandones, line 236: as above

etl/src/main/java/org/pih/hivmigration/etl/sql/HivExamOisMigrator.groovy
brandones: `source_patient_id` is interpolated from `source_encounter_id` by `migrate_tmp_obs`. I didn't realize that about TEXT and VARCHAR! I'll see if I can improve on the mapping. Thanks for the feedback.

brandones (approved):
A couple suggestions, but otherwise looks good!

etl/src/main/java/org/pih/hivmigration/etl/sql/HivExamOisMigrator.groovy
cioan: That's fine, but to be correct it should be:
m.openmrs_concept_source !='' and m.openmrs_concept_code != '';

cioan:
Thanks @brandones !

etl/src/main/java/org/pih/hivmigration/etl/sql/HivExamOisMigrator.groovy
cioan: You could use DATE_LENGTH to determine the length of the defined VARCHAR in the Oracle tables:

select COLUMN_NAME, DATA_TYPE, DATA_LENGTH from ALL_TAB_COLUMNS where TABLE_NAME='HIV_TB_STATUS';

or you could dynamically determine the length of the longest row in a given table:
select max(length(comments)) from HIV_EXAM_OIS;

PR #131, opened 2020-12-22
UHM-5044, concatenate HEENT normal obs to value_text

cioan:
@mogoodrich , I got this HEENT normal obs concatenated into the value_text.


mogoodrich (approved):
LGTM, thanks @cioan !

PR #132, opened 2020-12-23
UHM-5097 Migrate HIV_SOCIOECOMICS (part 2) and HIV_SOCOECONOMICS_EXTRA

brandones:
https://pihemr.atlassian.net/browse/UHM-5097

This is sort of a big one. I refactored `migrate_tmp_obs`. Now you can specify either `source_encounter_id` or `encounter_id` for obs. Also, the `source_patient_id` field is now ignored. This makes it easier to create obs for which the encounter does not correspond to anything in `hivmigration_encounters`.

I also refactored the RegistrationMigration, reducing its run time from ~10 mins to ~3 mins (for `-l=100000`).

### DIFF
--- a/etl/src/main/java/org/pih/hivmigration/etl/sql/LabResultMigrator.groovy
+++ b/etl/src/main/java/org/pih/hivmigration/etl/sql/LabResultMigrator.groovy
@@ -7,7 +7,6 @@ class LabResultMigrator extends ObsMigrator {
         executeMysql("Create staging table", '''
             create table hivmigration_lab_results (
               obs_id int PRIMARY KEY AUTO_INCREMENT,
-              source_patient_id int,
               source_encounter_id int,
               source_result_id int,
               sample_id VARCHAR(20),
@@ -26,8 +25,7 @@ class LabResultMigrator extends ObsMigrator {
 
         loadFromOracleToMySql('''
             insert into hivmigration_lab_results
-               (source_patient_id,
-                source_encounter_id,
+               (source_encounter_id,
                 source_result_id,
                 sample_id,
                 test_type,
@@ -37,10 +35,9 @@ class LabResultMigrator extends ObsMigrator {
                 value_boolean,
                 vl_beyond_detectable_limit,
                 vl_detectable_lower_limit)
-            values (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
+            values (?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
         ''', '''
-            select   e.patient_id as source_patient_id,
-                     e.ENCOUNTER_ID as source_encounter_id, 
+            select   e.ENCOUNTER_ID as source_encounter_id, 
                      r.RESULT_ID as source_result_id,
                      r.sample_id, 
                      lower(t.name) as test_type,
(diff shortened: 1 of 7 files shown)


etl/src/main/java/org/pih/hivmigration/etl/sql/LabResultMigrator.groovy
```diff
@@ -7,7 +7,6 @@ class LabResultMigrator extends ObsMigrator {
         executeMysql("Create staging table", '''
             create table hivmigration_lab_results (
               obs_id int PRIMARY KEY AUTO_INCREMENT,
-              source_patient_id int,
```
brandones, line 10: I did this for some migrators, but not all of them. Hence leaving `source_encounter_id` in the tmp_obs table for backwards-compatibility.

etl/src/main/java/org/pih/hivmigration/etl/sql/Migrator.java
```diff
@@ -122,12 +123,13 @@ public void main() {
                 migrate(new StagingTablesMigrator(), limit);
                 migrate(new PatientMigrator(), limit);
                 migrate(new InfantMigrator(), limit);
+                migrate(new EncounterMigrator(), limit);
```
brandones, line 126: RegistrationMigrator now depends on EncounterMigrator

etl/src/main/java/org/pih/hivmigration/etl/sql/RegistrationMigrator.groovy
```diff
@@ -162,181 +101,122 @@ class RegistrationMigrator extends SqlMigrator {
                     r.creator,
                     1    
             from hivmigration_registration_encounters r, hivmigration_patients p
-            where r.source_patient_id = p.source_patient_id;                                                                                      
+            where r.source_patient_id = p.source_patient_id;
+            
+            update hivmigration_patient_birthplace b
+            join hivmigration_patients hp on b.source_patient_id = hp.source_patient_id
+            join encounter e on hp.person_id = e.patient_id and e.encounter_type = @encounter_type_registration
+            set b.encounter_id = e.encounter_id;
```
brandones, line 109: This removes the need for all the joins later.

mogoodrich (commented):
@brandones  @cioan looks good to me have a quick review, but I'll defer to you guys as you've been more deeply into this code.

A few points/questions @brandones :
* Will this require us to re-test anything in detail? Registration?
* What's the use case for "This makes it easier to create obs for which the encounter does not correspond to anything in hivmigration_encounters."

brandones:
> Will this require us to re-test anything in detail? Registration?

I think that would be wise, unfortunately.

> What's the use case for "This makes it easier to create obs for which the encounter does not correspond to anything in hivmigration_encounters."

For example, there is no legacy socioeconomics form nor encounter. So the encounter is being created in the `encounter` table. These encounters have no corresponding encounter in `hivmigration_encounters`.

mogoodrich:
Makes sense @brandones ... can you make a ticket for what we will need to re-test?

brandones:
I looked and it's really just birthplace. I added a comment to this ticket.

brandones:
@mogoodrich Should I go ahead and merge?

mogoodrich:
Works for me, thanks @brandones !

PR #133, opened 2020-12-23
Part 2 of UHM-5090 Migrate current_tx.tb_other observation

brandones:
https://pihemr.atlassian.net/browse/UHM-5090

Addresses Mark's feedback:

> I agree with not mapping “Retraitement 2RHEZ/4RH” to “initial”, but shouldn’t it be mapped to “Retreatment: 2HRZE / 4HR“ (see patient 104145, followup encounter on Oct 30, 2019)
>
> I’m not seeing the mapping working properly for “2HZRE + 4 HR”, see patient 96440, encounter on Nov 4, 2020, or patient 94434, encounter on Oct 28, 2016… (this is specifically for the value with the spaces between “2HZRE” and “+” and “4” and “HR”… the others appear to be working)
>
> For the free-text values that can’t be mapped, adding a free text field on the new form is a bigger discussion, but I think as part of this ticket we should at least set the “anti-TB treatment” radio set to have “yes” checked… as this makes it clearer that the patient is at least on TB treatment


mogoodrich (approved):
LGTM, thanks @brandones !

PR #134, opened 2021-01-04
Ensure that multiple obs for 'currently on tb treatment' are not created

brandones:
https://pihemr.atlassian.net/browse/UHM-5090

Resolves
> For the free-text values that can’t be mapped, adding a free text field on the new form is a bigger discussion, but I think as part of this ticket we should at least set the “anti-TB treatment” radio set to have “yes” checked… as this makes it clearer that the patient is at least on TB treatment

So that if both a coded and non-coded treatment are provided, the coded treatment is the one that sets the "anti-TB treatment" radio.


mogoodrich (approved):
What was the issue, there was a coded treatment behind the scenes, with a value of null?

brandones:
With a value of "none", yes. In ~400 cases. Thanks @mogoodrich !

PR #135, opened 2021-01-04
Use datetime instead of timestamp for patient_created_date

brandones:
Should resolve https://pihemr.atlassian.net/browse/UHM-5075

The problem was actually upstream: the registration encounter_date was being set to the patient_created_date, but the patient_created_date wasn't being loaded correctly from the HIV_DEMOGRAPHICS table. All patients currently have `created_date` equal to the migration date.

I'll merge once I do an end-to-end run.

### DIFF
--- a/etl/src/main/java/org/pih/hivmigration/etl/sql/PatientMigrator.groovy
+++ b/etl/src/main/java/org/pih/hivmigration/etl/sql/PatientMigrator.groovy
@@ -25,7 +25,7 @@ class PatientMigrator extends SqlMigrator {
                 birth_place varchar(255),
                 accompagnateur_name varchar(255),
                 patient_created_by int,
-                patient_created_date timestamp,
+                patient_created_date datetime,
                 outcome varchar(255),
                 outcome_date date,
                 zone varchar(32),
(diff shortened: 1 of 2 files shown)


cioan (approved):
Thanks @brandones !

mogoodrich (commented):
LGTM... any takeaways here?  was siomething failing silently when we used a timestamp instead of a datetime?

brandones:
Yeah. `timestamp` only makes sense for a "last changed" field. Though this isn't obvious from the documentation. The trick is here in [this page](https://dev.mysql.com/doc/refman/5.6/en/timestamp-initialization.html):

> TIMESTAMP and DATETIME columns have no automatic properties unless they are specified explicitly, with this exception: If the explicit_defaults_for_timestamp system variable is disabled, the first TIMESTAMP column has both DEFAULT CURRENT_TIMESTAMP and ON UPDATE CURRENT_TIMESTAMP if neither is specified explicitly.

[explicit_defaults_for_timestamp](https://dev.mysql.com/doc/refman/5.6/en/server-system-variables.html#sysvar_explicit_defaults_for_timestamp) is disabled by default. Those automatic properties cause it to always be set to the current timestamp. Which is why it was ignoring the data we were trying to put into it.

PR #137, opened 2021-01-07
Default null encounter_date only after logging warning

brandones:
See https://github.com/PIH/hivmigration/pull/135


mogoodrich (approved):
Ah, okay... so we were already logging this error (like I thought) but it was causing downstream errors so you fixed it, and this commit just moves the fix until after the logging?  LGTM!

PR #141, opened 2021-01-08
UHM-5102 Migrate Follow-Up Side Effects section 

brandones:
https://pihemr.atlassian.net/browse/UHM-5102


cioan (approved):
Thanks @brandones !

PR #146, opened 2021-01-15
UHM-5096 Migrate HIV_CONTACTS table 

brandones:
https://pihemr.atlassian.net/browse/UHM-5096

### DIFF
--- a/etl/src/main/java/org/pih/hivmigration/etl/sql/ContactsMigrator.groovy
+++ b/etl/src/main/java/org/pih/hivmigration/etl/sql/ContactsMigrator.groovy
@@ -33,11 +33,11 @@ class ContactsMigrator extends ObsMigrator {
                    relationship,
                    CASE
                         WHEN cleaned_relationship IN ('fille', 'fils', 'enfant', 'enf', 'enfants', 'fiile', 'garcon', 'child', 'bellefille')
-                            THEN concept_uuid_from_mapping('PIH', 'CHILD')
+                            THEN concept_uuid_from_mapping('CIEL', '1528')  -- Child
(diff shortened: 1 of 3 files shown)


etl/src/main/java/org/pih/hivmigration/etl/sql/ContactsMigrator.groovy
```diff
@@ -0,0 +1,127 @@
...
+                hiv_status_uuid VARCHAR(38)
+            );
+        ''')
+
+        setAutoIncrement("hivmigration_tmp_contacts", "(select max(obs_id)+1 from obs)")
+
+        executeMysql("Populate temp table of patient contacts", '''
+            INSERT INTO hivmigration_tmp_contacts
+                (source_encounter_id, first_name, last_name, old_relationship, relationship_uuid, age, death_uuid, hiv_status_uuid)
+            SELECT he.source_encounter_id,
```
mseaton, line 30: See my comment on the ticket @brandones .  We should do a distinct query of contacts here, grouping on patient_id, in order to eliminate duplicate rows of contacts, and to see what the true maximum we need to support on the form is.

mogoodrich: Seems to make sense... @brandones is there thoughts on why this might not work?

brandones: Oh, it will be fine. We won't worry about the cases where there are different spellings (they will appear as separate contacts). We can do `MAX` on all the other fields, and if there's a discrepancy we can issue a data warning about it.

brandones:
@mogoodrich @mseaton @cioan Finally got deduplication and discrepancy warning working correctly. Ready for re-review.

mogoodrich (approved):
LGTM, thanks @brandones

PR #147, opened 2021-01-20
Log warning for patients with multiple intake encounters

brandones:
Turns out there's a number of patients (though not too many) with multiple intake encounters. Anywhere that we join on intake encounters, we may be producing duplicate obs (though perhaps that's okay).


mogoodrich:
LGTM thanks @brandones ... are this generally on the same day?  I just created a ticket about a (possibly) related issue I've been running into:

https://pihemr.atlassian.net/browse/UHM-5181

brandones:
Some are, some aren't. I've added the dates to the warning details.

PR #149, opened 2021-01-21
UHM-5095, make Sample ID top level obs

cioan:
I made Sample ID the top level obs for all lab results. I was not sure if I should remove the sample_id from each obs accession_number field. It seems harmless for now. What do you guys think? Thanks!


mogoodrich (approved):
LGTM, and, agreed, seems to make sense to keep the accession number as well...

cioan:
Thanks @mogoodrich !

PR #151, opened 2021-01-21
UHM-5187 - migrate regime hiv_encounters into visitless openmrs encou…

mseaton:
…nters

### DIFF
--- a/etl/src/main/java/org/pih/hivmigration/etl/sql/SqlMigrator.java
+++ b/etl/src/main/java/org/pih/hivmigration/etl/sql/SqlMigrator.java
@@ -1,17 +1,5 @@
 package org.pih.hivmigration.etl.sql;
 
-import org.apache.commons.dbutils.DbUtils;
-import org.apache.commons.dbutils.QueryRunner;
-import org.apache.commons.dbutils.ResultSetHandler;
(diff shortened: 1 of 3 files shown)


etl/src/main/java/org/pih/hivmigration/etl/sql/SqlMigrator.java
```diff
@@ -24,9 +12,20 @@
...
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.pih.hivmigration.etl.sql.util.FileParser;
+import org.pih.hivmigration.etl.sql.util.SqlStatementParser;
+import org.springframework.util.StringUtils;
+
 abstract class SqlMigrator {
 
-    private static final Log log = LogFactory.getLog(SqlMigrator.class);
+    private final Log log = LogFactory.getLog(getClass());
```
mseaton, line 28: FYI - this change makes it such that the log file shows the actual Migrator class that's running and spitting out messages, rather than always just showing SqlMigrator.  This makes it much easier to see where in the migration things are happening as you scroll / watch the log.

mogoodrich (approved):
LGTM!

cioan (approved):
Thanks!

PR #153, opened 2021-01-25
UHM-4867

mseaton:


mogoodrich (approved):
LGTM! I didn't go into the weeds, but that will come out in the testing!

mseaton:
@brandones / @cioan I'll look to merge this in later today, but wanted to give you a chance to have a look first in case you are interested / have feedback.

cioan (approved):
LGTM! Thanks!

etl/src/main/java/org/pih/hivmigration/etl/sql/SqlMigrator.java
```diff
@@ -317,4 +321,68 @@ private String abbreviate(String str) {
         }
     }
 
+    // Methods that allow for creating tests that should pass
+
+    protected void assertAllRows(String description, String mysqlQuery) throws Exception {
+        log.info("Testing: " + description);
+        Number allRows = (Number) selectMysql("select count(*) from drug_order", new ScalarHandler<Number>());
```
brandones, line 328: Looks like this function is specific to the `drug_order` table. Can you generalize it, maybe with a `tableName` parameter?

etl/src/main/java/org/pih/hivmigration/etl/sql/SqlMigrator.java
```diff
@@ -317,4 +321,68 @@ private String abbreviate(String str) {
...
+            for (String key : oracleRow.keySet()) {
+                allKeys.add(key.toLowerCase());
+            }
+            for (String key : mysqlRow.keySet()) {
+                allKeys.add(key.toLowerCase());
+            }
+            for (String key : allKeys) {
+                Object oracleValue = oracleRow.get(key.toUpperCase());
+                Object mysqlValue = mysqlRow.get(key);
+                if (oracleValue != null || mysqlValue != null) {
```
brandones, line 368: Can remove this outer `if` block and just make that last `else` into an `else if(mysqlValue == null && oracleValue == null)`.

brandones (approved):
Awesome work. One bug and one suggestion.

PR #157, opened 2021-01-28
UHM-5112: For Intake and Followup Forms, migrate Form Version

mogoodrich:


cioan (approved):

brandones (approved):

etl/src/main/java/org/pih/hivmigration/etl/sql/EncounterMigrator.groovy
```diff
@@ -54,16 +55,21 @@ class EncounterMigrator extends SqlMigrator {
...
+                    where e.patient_id = p.patient_id and i.ENCOUNTER_ID (+)= e.ENCOUNTER_ID and f.ENCOUNTER_ID (+) = e.ENCOUNTER_ID
             '''
         )
         executeMysql("Add UUIDs", '''
             UPDATE hivmigration_encounters SET encounter_uuid = uuid();
         ''')
 
+        executeMysql("Set Form Version to 1 for all Intake and Followup with no versions", '''
+            update hivmigration_encounters set form_version='1' where (source_encounter_type='intake' or source_encounter_type='followup') and form_version is null;
+        ''')
```
mseaton, line 71: You could use the coalesce function instead of the nvl function above to achieve this same thing I believe - eg. coalesce(i.form_version, f.form_version, 1) as form_version.  Just FYI.

mseaton (approved):
LGTM

etl/src/main/java/org/pih/hivmigration/etl/sql/EncounterMigrator.groovy
mogoodrich: Oh, right, I may try that, thanks!

PR #160, opened 2021-01-30
Change the validator functions in sql migrator to log data warnings r…

mseaton:
…ather than throw an Exception


mogoodrich:
Cool. Thanks for the quick response.  Will merge and rerun now to make sure everything works.

Good to have these warnings, and happy to work out some way for them to not get lost in the shuffle...

PR #161, opened 2021-02-01
Fix socioeconomic assistance migrator

brandones:


mogoodrich (approved):
LGTM... what was the change here?

brandones:
There were two problems. 1) The Nutritional and Transport aid questions were being migrated with the concept and value parameters swapped, and 2) An Other Aid obs was being added for each other type of aid, rather than concatenating them.

PR #164, opened 2021-02-04
UHM-5121, set hivmigration users as retired

cioan:


mogoodrich (approved):
LGTM, thanks @cioan !

PR #169, opened 2021-02-12
UHM-5108 pt 5: Migrate socioecon assist rows when comment = null and form_version < 3

brandones:
Also restores Nutritional Aid migration, which was lost somehow.


mogoodrich (approved):
LGTM

cioan (approved):

PR #171, opened 2021-02-13
UHM-5186 Add Key Population sample data 

brandones:
https://pihemr.atlassian.net/browse/UHM-5186

My database has 42754 patients in hivmigration_patients. That means there should be 213770 entries. This generates 201840. I don't know how to account for the discrepancy but it's pretty close, and I figured I'd submit this for review.

It also takes a long time. 1.5 hours on my machine. Most of that time is spent filling in the "unknown" and "no" answers.


brandones:
Removed the part that fills in "no" answers and it's now running in a few minutes!

brandones:
Shall we call this good since Debbie took a look, @mogoodrich ?

mogoodrich:
Works for me, thanks @brandones !

PR #172, opened 2021-02-16
UHM-5124, migrate only obs with form_version 3

cioan:


etl/src/main/java/org/pih/hivmigration/etl/sql/HivExamOisMigrator.groovy
```diff
@@ -99,15 +102,29 @@ class HivExamOisMigrator extends ObsMigrator {
...
+                end as form_name,
+                case
+                    when (x.encounter_id in (select k.encounter_id from HIV_INTAKE_FORMS k where k.encounter_id=x.encounter_id)) then 
+                    ( select form_version from HIV_INTAKE_FORMS where encounter_id=x.encounter_id ) 
+                    when (x.encounter_id in (select f.encounter_id from HIV_FOLLOWUP_FORMS f where f.encounter_id=x.encounter_id)) then 
+                    ( select form_version from HIV_FOLLOWUP_FORMS where encounter_id=x.encounter_id ) 
+                    else null 
+                end as form_version
             from HIV_EXAM_OIS x, HIV_ENCOUNTERS e, hiv_demographics_real d  
             where x.ENCOUNTER_ID = e.ENCOUNTER_ID and e.patient_id = d.patient_id;
```
brandones, line 129: Any reason you didn't just `JOIN` on the `_FORMS` tables? Would simplify the code and might be faster.

brandones (approved):
Looks generally good to me, just one suggestion/note

etl/src/main/java/org/pih/hivmigration/etl/sql/HivExamOisMigrator.groovy
mogoodrich: Yeah, I think a join would work here?

cioan: Thanks guys! I have updated the query to use joins. Thanks!

PR #175, opened 2021-02-18
UHM-5110 Migrate/Fix Migration of Intake Prophylaxis section (20)

brandones:
https://pihemr.atlassian.net/browse/UHM-5110


mogoodrich:
LGTM, though it might be worth taking a look at my recent comments on the PR I just reverted since I see the hivmigration_ordered_other table is in play here as well...

PR #181, opened 2021-02-24
Add index to tmp_obs.obs_group_id supporting the query in TreatmentObs:migrateArtStatus - 'Add codifiable values...'

brandones:
Speeds up "Add codifiable values that were put in 'other', if there's not already a coded value." See [the last build](http://bamboo.pih-emr.org:8085/download/HIVM-EMJ-JOB1/build_logs/HIVM-EMJ-JOB1-256.log), where that one query took 50 mins.


mogoodrich (commented):
awesome, thanks @brandones !

PR #184, opened 2021-02-26
UHM-5106 Fix migration of Ordered ARV - handle case where both ARV #1 and #2 are codifiable

brandones:
https://pihemr.atlassian.net/browse/UHM-5106


mogoodrich (approved):
LGTM thanks @brandones !

PR #186, opened 2021-03-01
UHM-5066, migrate only symptoms captured on intake v3 form

cioan:
Hi @mogoodrich , here is the update to only migrate symptoms captured on v3 intake form. Thanks!


cioan:
Thanks @mogoodrich !

PR #194, opened 2021-03-04
UHM-5261, map null location names to Unknown Location

cioan:


mogoodrich (commented):
@cioan what is the change here?  

It looks like we were already mapping those two locations to "Unknown Location", so maybe this ticket was a mistake on my part?

cioan:
I just checked haitihivtest and confirm that those 2 null locations are already mapped to the Unknown Location.

mogoodrich:
Strange!  So no need to change this then... I will assign this back to myself for analysis...

PR #196, opened 2021-03-05
UHM-5106 Fix migration of Ordered ARV - don't create obs with null va…

brandones:
…lue_text


mogoodrich (approved):
LGTM, thanks @brandones !

PR #197, opened 2021-03-08
UHM-5270, migrate radiology exams

cioan:


mogoodrich:
LGTM, thanks @cioan !

PR #198, opened 2021-03-08
UHM-5261: Handle HIV_INSTITUTIONS with NAME = null

mogoodrich:


etl/src/main/java/org/pih/hivmigration/etl/sql/ExamExtraMigrator.groovy
```diff
@@ -350,7 +350,7 @@ class ExamExtraMigrator extends ObsMigrator {
                 source_encounter_id,
                 concept_uuid_from_mapping('PIH', '13170') as concept_uuid
             FROM hivmigration_transfer_out_to
-            WHERE transfer_out_to is not null;
```
mogoodrich, line 353: So the transfer locations for these two locations weren't getting migrated because athough the hiv_institution_id wasn't null, the transfer_out_of value (which was just the text name of the institution) was not.

cioan (approved):
Thanks @mogoodrich !

mogoodrich:
Thanks @cioan !  Merging this in....

PR #200, opened 2021-03-10
UHM-5258, adjust enrollment date if it is after the completion date

cioan:


mogoodrich (approved):
LGTM, thanks @cioan !

PR #202, opened 2021-03-10
UHM-5293, update infants dead status

cioan:


mogoodrich (approved):
LGTM, thanks @cioan !

PR #204, opened 2021-03-11
UHM-5268, migrate pcr_test_id field to hivmigration_pcr_tests

cioan:
@mogoodrich , I have replaced the HIV_PCR_TRACKING_VIEW with the entire select statement behind the view, and I just added the pcr_test_id. Thanks!


mogoodrich (approved):
Makes sense to me!  I'll add a comment to remember to a spot-check of the PCR_TESTs migration when testing this ticket, to make sure it's still works as expected.

PR #206, opened 2021-03-11
UHM-5296, migrate encounter's comments

cioan:


mogoodrich (approved):
LGTM, though not sure if the HivExamOisMigrator is the right place for this to go?  Maybe directly in the EncounterMigrator?  Or is there a specific reason you put it hear that I'm not thinking of.

cioan:
Thanks @mogoodrich ! I have added it here because the obs in that form are in this migrator and it was easier to just create another obs in here. The Encounter migrator does not extend ObsMigrator, so it would require more gymnastics to get that obs created in there.

PR #208, opened 2021-03-19
Tb status migrator

mseaton:


etl/src/main/java/org/pih/hivmigration/etl/sql/TbStatusMigrator.groovy
```diff
@@ -0,0 +1,223 @@
...
+                          ) s on i.source_patient_id = s.source_patient_id
+            set           i.STATUS_DATE = s.first_status_date, 
+                          i.PPD_POSITIVE_P = if(num_ppd_positive > 0, 't', 'f'), 
+                          i.TB_ACTIVE_P = if(num_tb_active > 0, 't', 'f'), 
+                          i.EXTRAPULMONARY_P = if(num_extrapulmonary > 0, 't', 'f'), 
+                          i.PULMONARY_P = if(num_pulmonary > 0, 't', 'f'), 
+                          i.DRUG_RESISTANT_P = if(num_drug_resistant > 0, 't', 'f'), 
+                          i.DRUG_RESISTANT_COMMENT = s.all_drug_resistant_comment
+            where         i.status_on_encounter = false
+            ;
```
mogoodrich, line 102: Is the assumption here that all statuses not linked to an encounter came from an Intake, not a follow-up?  There are about 500,000+ rows in the TB Status table linked to follow-up encounters.

mogoodrich (commented):
One question @mseaton ... also wondering re: follow-up encounters... are we assuming we don't need to migrate those at all?

etl/src/main/java/org/pih/hivmigration/etl/sql/TbStatusMigrator.groovy
mseaton: No.  It doesn't make that assumption.  It's just trying to establish if the unlinked data could presumably be migrated in for the intake encounter.

mogoodrich: Maybe I'm not following, but it seems to be making that assumption... or, I guess to be clearer, it's saying "if we don't have a linked Intake encounter for a patient, assume that any row not linked to an encounter can be used to populate the Intake in OpenMRS", because this query is just pulling all the rows in TB status without an encounter and then linking them to to the hivmigration_tb_status_intake for all rows/patients that haven't been populated by a linked intake encounter?

mseaton: Yes, I think you are right.  We can debate this.  It's possible we want to change this, but i'm not entirely sure the best way to do so.

mogoodrich: It's possible that that assumption may be fine... do we know the history of statuses that aren't linked to encounters?  

We also could skip this part entirely has legacy data from v1/v2/not MVP if we know that all the v3 Intakes are linked to statuses by encounter_id, so we have those covered.

PR #209, opened 2021-03-19
UHM-5310 Fix migration of ARV start date on Intake and Followup

brandones:

### DIFF
--- a/etl/src/main/java/org/pih/hivmigration/etl/sql/TreatmentObsMigrator.groovy
+++ b/etl/src/main/java/org/pih/hivmigration/etl/sql/TreatmentObsMigrator.groovy
@@ -322,14 +322,14 @@ class TreatmentObsMigrator extends ObsMigrator {
             INSERT INTO hivmigration_tmp_arv_regimen (source_encounter_id, coded, other)
             SELECT hoo1.source_encounter_id, trim(hoo1.value), trim(hoo2.value)
             FROM hivmigration_observations hoo1
-                     LEFT JOIN hivmigration_observations hoo2 ON hoo1.source_encounter_id = hoo2.source_encounter_id AND hoo2.observation = 'current_tx.art_other'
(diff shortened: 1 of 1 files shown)


etl/src/main/java/org/pih/hivmigration/etl/sql/TreatmentObsMigrator.groovy
```diff
@@ -753,6 +750,7 @@ class TreatmentObsMigrator extends ObsMigrator {
                             obs_group_id
                      FROM hivmigration_tmp_arv_regimen) arv
             LEFT JOIN tmp_obs o ON o.obs_group_id = arv.obs_group_id
+                AND o.concept_uuid = concept_uuid_from_mapping('PIH', '1282')
```
brandones, line 753: This query says "make a temp table of coded mappings for non-coded values. Then join on the existing obs to filter down to just the groups which don't already have a coded obs value." The problem was that it wasn't checking "whether there's already a coded obs value in the group," it was checking "whether there's anything in the group," which there always ways.

cioan (approved):

etl/src/main/java/org/pih/hivmigration/etl/sql/TreatmentObsMigrator.groovy
mogoodrich: Just to clarify, Is this for the issue with the art start date not getting migrated over if a regimen wasn't specified from my last comment on the ticket... ?  This changes seems unrelated but maybe I'm missing something?

brandones: Since this ART entry wasn't getting added, "Autre" was getting filled in for the coded medication. The date seems to be working, at least in the follow-up status section. It's even working for non-codable other values:

Screenshot from 2021-03-19 11-57-37

PR #212, opened 2021-03-23
UHM-5110 Migrate/Fix Migration of Intake Prophylaxis section (20) again

brandones:
https://pihemr.atlassian.net/browse/UHM-5110

Screenshot from 2021-03-23 11-29-17

### DIFF
--- a/etl/src/main/java/org/pih/hivmigration/etl/sql/TreatmentObsMigrator.groovy
+++ b/etl/src/main/java/org/pih/hivmigration/etl/sql/TreatmentObsMigrator.groovy
@@ -11,14 +11,14 @@ class TreatmentObsMigrator extends ObsMigrator {
         // the hivmigration_observations table gets created by ObsLoadingMigrator
         // the hivmigration_ordered_other table gets created by StagingTablesMigrator
 
-//        migrateProphylaxesState()
+        migrateProphylaxesState()
         migrateProphylaxesPlan()
-//        migrateArtStatus()
-//        migrateArtPlan()
-//        migrateTbState()
-//        migrateTbPlan()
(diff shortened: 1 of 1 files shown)


etl/src/main/java/org/pih/hivmigration/etl/sql/TreatmentObsMigrator.groovy
```diff
@@ -136,10 +136,11 @@ class TreatmentObsMigrator extends ObsMigrator {
     }
 
     def void migrateProphylaxesPlan() {
+        // This is actually historical information.
+        // But it's in the Plan section on the OpenMRS Intake form for some reason.
```
mogoodrich, line 140: Why are you assuming this is historical information?  The history or prophylaxes is collected in section 7 of the HIV EMR Intake form (as have already been migrated), this is about migrating the "plan" information, which is collected in Section 20 of the HIV EMR Intake.

What suggests to you the fact that this is historical?  The "stop" and "continue".  I would assume that it's possible that patients are already on prophylaxes upon intake, but this is still a "plan" whether or not to continue that treatment?

brandones: Oh, that's a fair point. It's "stopped" and "continued" (in the past tense), which is what got me mixed up. I think I've spent too much time with this piece of form and it's made my head spin.

mogoodrich: That's fair enough re: head spinning! :)

mogoodrich (approved):

PR #222, opened 2021-03-29
Fix UHM-5292 Migrate other data points from HIV_INFANTS table into OVC Intake encounter

brandones:
Fix for https://pihemr.atlassian.net/browse/UHM-5292

### DIFF
--- a/etl/src/main/java/org/pih/hivmigration/etl/sql/ObsMigrator.groovy
+++ b/etl/src/main/java/org/pih/hivmigration/etl/sql/ObsMigrator.groovy
@@ -32,11 +32,15 @@ abstract class ObsMigrator extends SqlMigrator {
     }
 
     void migrate_tmp_obs() {
-        executeMysql("Prepare tmp_obs table for migration", '''
-            UPDATE tmp_obs
-                LEFT JOIN hivmigration_encounters he ON tmp_obs.source_encounter_id = he.source_encounter_id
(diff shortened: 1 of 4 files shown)


etl/src/main/java/org/pih/hivmigration/etl/sql/ObsMigrator.groovy
```diff
@@ -32,11 +32,15 @@ abstract class ObsMigrator extends SqlMigrator {
...
-            WHERE tmp_obs.encounter_id IS NULL;
+        if (tableExists("hivmigration_encounters")) {
+            executeMysql("Add encounter id from hivmigration_encounters", '''
+                UPDATE tmp_obs
+                    LEFT JOIN hivmigration_encounters he ON tmp_obs.source_encounter_id = he.source_encounter_id
+                SET tmp_obs.encounter_id = he.encounter_id
+                WHERE tmp_obs.encounter_id IS NULL;
+            ''')
+        }
+        executeMysql("Add encounter datetime if not present", '''
```
brandones, line 43: This is the only change from the last time.

cioan: Are there any circumstances when we need to create obs without an encounter? I do not understand why we need to check for the existence of hivmigration_encounters table?

cioan (approved):
Thanks @brandones !

PR #224, opened 2021-03-31
UHM-5270, remove duplicate chest xRay constructs

cioan:
Good catch, @mogoodrich ! Thanks!


mogoodrich (approved):
Thanks @cioan , LGTM!

PR #227, opened 2021-04-01
UHM-5015, migrate treatment statuses

cioan:


mogoodrich (approved):
Thanks @cioan ... I didn't double-check every mapping, but reviewing the overall process and it looks good!

PR #228, opened 2021-04-01
UHM-5310: Fix query for ART and ART-other

brandones:


mogoodrich (approved):
LGTM... the only caveat being that both the "Art Status" and "Art Regimen" still use the same temp table (hivmigration_tmp_arv_regimen) which doesn't really matter in the long run, of course, since it is dropped and recreated each time, but it  meant when I was trying to test "Art Status" and found discrepancies I couldn't look back at the temp table (since it now contained the Art Regimen information).

PR #232, opened 2021-04-05
UHM-5239 - Map new locations into location migrator

mseaton:


mogoodrich (approved):
Didn't confirm all the uuids are correct, but will check when testing... :)

PR #236, opened 2021-04-07
UHM-5015, update treatment status date

cioan:


mogoodrich (approved):
This looks right @cioan though it looks like it's showing the entire diff, not just the status date change... so assuming that's all that changed... :)

cioan:
That's correct, @mogoodrich ! Thanks!

PR #243, opened 2021-04-13
UHM-5395 - Add Data Warnings - future relationships

mseaton:


mogoodrich (approved):
I'm fine with this as-is, but do we want to add referenced to both persons in the relationship?

mseaton:
I considered it, but don't really see the value.  The point is to highlight the issue - doing it on the parent side of the relationship makes sense to me.  This is also likely to be completely overlapping with any warning around future birthdates, as the relationship date is set to the child's birthdate, which is what is ending up in the future.

PR #244, opened 2021-04-13
UHM-5412 - Update Encounter in Future and Nonsensical Encounter Date …

mseaton:
…validators to log encounter type in warning_details


mogoodrich (approved):
LGTM!

PR #264, opened 2022-02-28
UHM-6212 - encounters with Unknown Location for patients not enrolled

mseaton:


etl/src/main/java/org/pih/hivmigration/etl/sql/PatientMigrator.groovy
```diff
@@ -87,11 +89,22 @@ class PatientMigrator extends SqlMigrator {
...
                     d.TREATMENT_STATUS_DATE as outcome_date,
-                    d.zone as zone
+                    d.zone as zone,
+                    d.starting_health_center,
+                    d.health_center
                 from HIV_DEMOGRAPHICS_REAL d
                 order by  d.PATIENT_ID
             ''');
 
+        executeMysql("Remove bad health center entries", '''
```
mogoodrich, line 99: What are 0 and 42?  Keys that don't reference anything, or a bogus location?

mogoodrich (commented):
One question, but LGTM, thanks @mseaton !

etl/src/main/java/org/pih/hivmigration/etl/sql/PatientMigrator.groovy
mseaton: No idea how something would be 0, but 42 = "Other (non-ZL)" in the old system.  I added this because I found this in the ProgramMigrator:

```
        executeMysql("Remove bad health center entries", '''
            UPDATE hivmigration_programs_raw
            SET health_center = IF (health_center = 0 OR health_center = 42, NULL, health_center),
                starting_health_center = IF (starting_health_center = 0 OR starting_health_center = 42, NULL, starting_health_center)
            ;
        ''')
```
So, I was following this existing pattern.

PR #267, opened 2022-03-10
UHM-5432 - Default encounter dates to vital status date if birthdate …

mseaton:
…is null, add additional data warnings around encounter dates.


mogoodrich (approved):
LGTM, thanks @mseaton !

